PR #63, opened 2014-03-13
Fixing a NoneType is not callable error

croach:
If a QuerySelectionField object has a query attribute that is considered by Python to be False (e.g., an empty sequence or mapping object), and the query_factory attribute is None, an error occurs (NoneType is not callable error). To fix this, this change checks specifically for None on both the query and query_factory attributes and defaults to an empty list if both are None. This gives the user a little more control on when to execute the query_factory callable. With the previous code, the user would have to return a list with something in it if they did not want the query_factory to be called, now they can return an empty list if the field should be empty and a None if they want the query_factory to be called when the query is empty.

### DIFF
--- a/wtforms/ext/sqlalchemy/fields.py
+++ b/wtforms/ext/sqlalchemy/fields.py
@@ -95,7 +95,11 @@ class QuerySelectField(SelectFieldBase):
 
     def _get_object_list(self):
         if self._object_list is None:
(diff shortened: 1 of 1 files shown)


crast:
Closing because `wtforms.ext.sqlalchemy` is no longer part of wtforms

PR #81, opened 2014-05-06
Adding custom attributes for option tags within SelectField and SelectMultipleField

playpianolikewoah:
Currently you do not allow the user to create html attributes on `<option>` tags within `SelectField` and `SelectMultipleField`. For example if a user wanted `<option data-id="3">` it is not possible currently.

### DIFF
--- a/wtforms/widgets/core.py
+++ b/wtforms/widgets/core.py
@@ -280,7 +280,7 @@ class Select(object):
             options = {}
         html = ['<select %s>' % html_params(name=field.name, **kwargs)]
         for val, label, selected in field.iter_choices():
-            html.append(self.render_option(val, label, selected,**options))
+            html.append(self.render_option(val, label, selected, **options))
         html.append('</select>')
         return HTMLString(''.join(html))
 


hjl-csis:
Did something change regarding this issue?
I get the workaround, but would like to keep my view clean and define all the choices in one place at the form initialization.
Will wtforms 3 change something in that regard? Thanks in advance.

dubbl:
Did something change regarding this issue?
I get the workaround, but would like to keep my view clean and define all the choices in one place at the form initialization.
Will wtforms 3 change something in that regard? Thanks in advance.

goodwordalchemy:
+1 for this. It's a big problem and makes the code confusing...

robinclark007:
Really need this, I find this via [stackoverlow](http://stackoverflow.com/questions/23460857/create-selectfield-options-with-custom-attributes-in-wtforms)

and the default only echo out a <option value='bb'>kk</options>
really need a way to costomize the option 
guess I have to use this unclean way...

so lame,,

trentkg:
Any updates on this?

roipoussiere:
How do I add an attribute then?

I tried this but it doesn't work:

```python
form = FormWithRadioButton()
for option in form.my_radio_button:
    setattr(option, 'title', 'hello')
```

Issue #48, opened 2014-01-27
General way to pass extra attributes to Field


yegle:
According to http://www.w3.org/TR/html-markup/syntax.html#syntax-attributes attribute name allows almost all characters, including `-` (dash)

I'm trying to create an `<input>` tag with attribute `data-toggle` to work with `Flat-UI`'s css definition, but unfortunately current design of `Field.__call__` doesn't allow such attribute names.

xcash:
@prencher @patrickyan You're right on the specs but let's get practical. How many users are using underscores in data attributes opposite to how many are using frameworks such as Bootstrap which uses  data-x-y-z format? 
If the majority of us are using the latter it's not nice to force all of us use the hacky `**` syntaxt.

This is a typical breaking change for a major release maybe. :)

PS: To preserve the opportunity to use underscore a solution may be to use an escape character which you can pass only using the `**` syntax.

What do you think?

crast:
Finished in #248

PR #114, opened 2014-08-04
Replace all underscores with hyphen if attribute name starts with 'data_...

jeffcjohnson:
...', not just the first underscore.

See Issue #48.


crast:
Going to accept https://github.com/wtforms/wtforms/pull/248 instead

Issue #120, opened 2014-08-23
Multiple labels supported?


devurandom:
For my form I need the fields to have multiple labels – one for each language I support. This is because I used a very simple approach to i18n [1], which works very well on my small website:
```css
body.en :lang(de) {
  display: none;
}
body.de :lang(en) {
  display: none;
}
```
```html
<select id="lang" onChange="document.body.className = this.value">
  <option value="en">English</option>
  <option value="de">Deutsch</option>
</select>
…
<label for="thing" lang="en">Thing</label>
<label for="thing" lang="de">Ding</label>
<input id="thing" name="thing" type="text"/>
```

Is this somehow possible with WTForms?

[1] http://stackoverflow.com/a/13607315/1975217

crast:
You can do this with no changes to WTForms:

```python
StringField(dict(de='German Label', en='English Label')
```

Then render in jinja as:
```jinja
<label for="{{ field.id }}">{{ field.label.text.de }}</label>
```

PR #121, opened 2014-08-24
Fix wtforms/wtforms#120: Allow custom Labels in Field's __init__

devurandom:
Fixes wtforms/wtforms#120 by allowing code like the following:

``` python
class i18nLabel(Label):
  def __init__(self, field_id, **kwargs):
    super(i18nLabel, self).__init__(field_id, None)
    self.translations = kwargs

  def __call__(self, text=None, **kwargs):
    if 'for_' in kwargs:
       kwargs['for'] = kwargs.pop('for_') 
    else:
       kwargs.setdefault('for', self.field_id)

    attributes = html_params(**kwargs)
    return ''.join(HTMLString('<label %s lang="%s">%s</label>' % (attributes, lang, text)) for lang, text in self.translations.iteritems())

_ = i18nLabel

class ApplicationForm(Form):
  name = StringField(_('name', de='German Label', en='English Label'))
```

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -5,7 +5,7 @@ import decimal
 import itertools
 
 from wtforms import widgets
-from wtforms.compat import text_type, izip
+from wtforms.compat import string_types, text_type, izip
 from wtforms.i18n import DummyTranslations
 from wtforms.validators import StopValidation
 from wtforms.utils import unset_value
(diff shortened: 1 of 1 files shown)


ftm:
This seems like quite a niche use case. It's been nearly 4 years and there are good ways around this restriction so I'm going to close this.

PR #131, opened 2014-10-06
Smarter missing input validation.

### LINKED ISSUE #100: InputRequired doesn't accept 0 as valid
Hello,

I have a form that looks like this,

    class VoteForm(Form):
        article_id = IntegerField('article_id', [validators.DataRequired()])
        vote_type  = IntegerField('vote_type', [validators.InputRequired(), validators.NumberRange(min = 0, max = 1)])

and create the form with,

    form = VoteForm(data = request.json_body)

`request.json_body` is the dictionary `{u'article_id': 1, u'vote_type': 0}`.

I would expect that the value 0 is fine with `InputRequired` but it seems it is not.

Guzzardo:
Resolves #100.

### DIFF
--- a/tests/validators.py
+++ b/tests/validators.py
@@ -129,11 +129,11 @@ class ValidatorsTest(TestCase):
 
     def test_input_required(self):
         self.assertEqual(input_required()(self.form, DummyField('foobar', raw_data=['foobar'])), None)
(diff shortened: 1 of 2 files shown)


crast:
This is a major breaking change and is probably not a good idea.

Issue #109, opened 2014-07-24
populate_obj creates dirty objects in session


snc-tolksdorf:
if I use

form.populate_obj(user)

and the user has "active = True" and the new value for active is True, the user object is marked as dirty in the session because of

/wtforms/fields/core.py :: Field

    def populate_obj(self, obj, name):
        """
        Populates `obj.<name>` with the field's data.

        :note: This is a destructive operation. If `obj.<name>` already exists,
               it will be overridden. Use with caution.
        """
        setattr(obj, name, self.data)

should it be like this?

    def populate_obj(self, obj, name):
        if hasattr(obj, name) and getattr(obj, name) != self.data:
            setattr(obj, name, self.data)

longhotsummer:
WTForms relies on the model to check if values have changed. For instance, SQLAlchemy does this for its models. You can call `model.active = True` as much as you like but SQLAlchemy will only write that change if it is different to what is in the database.

I don't think it should be the responsibility of WTForms to decide if a value has changed or not, that should be up to the model.

PR #147, opened 2014-12-22
Feature proposal: Partial object updates

codebynumbers:
I often find myself wanting to use wtforms to populate an object for PUT operations. However the current behavior of populate_obj() essentially clobbers all model properties if even if they were not intended to be changed. This would allow the use of wtforms to validate fields and partially apply them to an object.


hnb2:
I am looking forward to see this PR merged, it will allow the use of patched data. And it will solve this ticket too I guess: https://github.com/wtforms/wtforms/issues/109

crast:
I'm shelving this as far future

debrice:
Does this also addresses the case where a field value would be required but not provided:
For example, a title field could be required only when provided, the form wouldn't fail if the "title" was not a part of `BaseForm._formkeys`

davidism:
It looks like this doesn't account for validation. It also can't account for nested fields, which the other PR had an issue with too.

POST and PUT both have the expectation that the full resource is sent. PATCH does not, but also implies a more sophisticated algorithm than implemented here.

I recognize that support for PATCH requests would be useful, but I think we need to do some more design work on this still.

Issue #148, opened 2014-12-29
Generalized whitespace stripping for form fields


tom-dignan:
I think it's an appropriate feature to add to WTForms. If you agree, where would the best place to add it be?

crast:
You can do it in WTForms 2.x by using the [bind_field](http://wtforms.readthedocs.org/en/2.0.2/meta.html#wtforms.meta.DefaultMeta.bind_field) primitive on `class Meta`

Because class Meta can be inherited, you can actually then set up your own new base form primitives:

```python
class MyBaseForm(Form):
    class Meta:
        def bind_field(self, form, unbound_field, options):
            filters = unbound_field.kwargs.pop('filters', None) or []
            filters.append(my_strip_filter)
            return unbound_field.bind(form=form, filters=filters, **options)
```

Now, just inherit MyBaseForm for all your forms and you're good to go.

ThiefMaster:
Does adding default filters somehow modify the process of unbound fields being bound to a form instance? I'm using a `QuerySelectMultipleField` where I'm assigning the `query` property after creating an instance of the form and as soon as there's a filter set - even if it's just `lambda x: x` - processing my form fails because `query` is None.

ThiefMaster:
Found another bug: popping the `filters` kwarg will break all but the first instance of the form if there were additional filters. Using `get` instead of `pop` fixes this.

pcraig3:
I've just been trying to implement this, but unfortunately, it doesn't seem to be working for `FieldList`s.  Binding filters is working quite nicely for regular fields, but trying to add them onto a `FieldList` throws the following error:
```
TypeError: FieldList does not accept any filters. Instead, define them on the enclosed field.
```

At the moment, the `bind_field` method of my subclassed `Form` looks like this:
```python
def bind_field(self, form, unbound_field, options):
            filters = unbound_field.kwargs.get('filters', [])
            if unbound_field.field_class is not FieldList:
                filters.append(strip_input)
            return unbound_field.bind(form=form, filters=filters, **options)
```

I know I could add the filter to the Field I'm passing into the FieldList (`FieldList(StringField('Name', filters=[some_filter]))`), but it seems to me that that somewhat defeats the purpose of subclassing the form to begin with.

What would be the problem with allowing filters passed in to a `FieldList` to be passed on to enclosed fields?  I'm thinking about changing [this line here](https://github.com/Khan/wtforms/blob/master/wtforms/fields/core.py#L834) to something like:
```python
[...]

PR #150, opened 2015-01-07
Add filters to forms

bootandy:
Now you can apply a filter to a form and it will be automatically added to all fields in that form


bootandy:
I've seen people asking for this, let me know if this is the correct way to do it.
https://github.com/wtforms/wtforms/issues/148
https://github.com/wtforms/wtforms/issues/128

PR #151, opened 2015-01-09
Add filters to forms

bootandy:
Now you can apply a filter to a form and it will be automatically added to all fields in that form

I've seen people asking for this, let me know if this is the correct way to do it.
https://github.com/wtforms/wtforms/issues/148
https://github.com/wtforms/wtforms/issues/128

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.pyo
 .*.swp
 .DS_Store
+.idea/*
 
 # Extra Files
 MANIFEST
--- a/tests/form.py
+++ b/tests/form.py
@@ -11,9 +11,10 @@ from tests.common import DummyPostData
 
 class BaseFormTest(TestCase):
     def strip_ws(self, s):
-        if isinstance(s, basestring):
-            s = s.strip()
-        return s
+        try:
+            return s.strip()
+        except AttributeError:
+            return s
 
     def get_form(self, **kwargs):
         def validate_test(form, field):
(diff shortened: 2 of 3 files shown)


crast:
This is a really simple feature, but the API guarantees on Form mean it will means this would be a feature that would cause a major version bump (to version 3.x) because it hides a possible field name on Form and changes `__init__`. Because of that, and the awkward timing right now, in that we are working on getting WTForms 3 stabilized which is a heavy re-think of many things, it's going to have to wait

Furthermore, it's made less urgent because  by extending [bind_field](http://wtforms.readthedocs.org/en/2.0.2/meta.html#wtforms.meta.DefaultMeta.bind_field) on `class Meta` you can achieve similar effect:

```python
class MyBaseForm(Form):
    class Meta:
        def bind_field(self, form, unbound_field, options):
            filters = unbound_field.kwargs.get('filters', None) or []
            filters.append(my_strip_filter)
            return unbound_field.bind(form=form, filters=filters, **options)
```

Now, just inherit MyBaseForm for all your forms and you're good to go.

erebus1:
Please, be careful, with this PR:

When you do filters.append(smth) then on each form initialization filters growth by 1 element. As a result, your code works slower and slower until you restart it

Consider Example:
```
   class MyBaseForm(Form):
        class Meta:
            def bind_field(self, form, unbound_field, options):
                filters = unbound_field.kwargs.get('filters', [])
                filters.append(my_strip_filter)
                return unbound_field.bind(form=form, filters=filters, **options)


    def my_strip_filter(value):
        if value is not None and hasattr(value, 'strip'):
            return value.strip()
        return value

    class MyCustomForm(MyBaseForm):
        some_field = StringField(filters=[lambda x: x])

    for i in range(100):
        MyCustomForm(MultiDict({'some_field': 'erer'}))

    print(len(MyCustomForm.some_field.kwargs['filters']))  # print: 101
```
So the fast fix is to check that this filter not in list:
```
class MyBaseForm(Form):
    class Meta:
        def bind_field(self, form, unbound_field, options):
            filters = unbound_field.kwargs.get('filters', [])
            if my_strip_filter not in filters:
                filters.append(my_strip_filter)
        return unbound_field.bind(form=form, filters=filters, **options)
```

davidism:
This can be handled in `Meta`.

PR #155, opened 2015-01-25
added Angular attribute support

jstacoder:
currently the only attributes supported with a '-' dash in the key is for data- attributes, this adds a small change to that to allow for angularjs ng- prefixed attributes as well, makes integrating with angular projects much less painful.


crast:
I'm a bit torn on this. My biggest problem is, ng- attributes are not in the HTML spec, while [data attributes are](http://www.w3.org/html/wg/drafts/html/master/dom.html#custom-data-attribute). The flip side, making developers lives easier is a great place to be.

jstacoder:
if we could get #114 passing the builds, that would work to. Because in angular attributes prefixed with `data-ng-` get changed to `ng-` internally, but in the current implementation of `html_params` only the first underscore is changed to a hyphen, so `data_ng_xxxx` would be `data-ng_xxxxx` and angular misses it.   
I really feel that a majority of people using wtforms could really benefit (if they used it) from the [amount of help with form/field validation/state management  provided by angular](http://www.ng-newsletter.com/posts/validations.html), but thats only possible if something is done. Because we know most people arent going to be able to discern that the small change i added is all that is needed to add support, which would add a ton of overhead to any other sort of form validation on the front end that ties the state of validation to   
* the css 
* input element 
* form 
* and surrounding elements

davidism:
I want to support this, but I don't want to add a check for `ng_` specifically since it's not a standard attribute and there may be any number of other frameworks that this pattern could apply to. I think a policy attribute on the `Meta` class, like `html_hyphenate_attributes={'data', 'aria', 'ng'}, might be a good solution.

davidism:
This can be accomplished by overriding a form's `Meta.render_field` method, so I don't think I'll make any other changes for now.

```python
class NgForm(Form):
    class Meta:
            for key, value in render_kw.items():
                if value.startswith('ng_'):
                    render_kw[key] = value.replace('_', '-')
            super().render_field(field, render_kw)

PR #156, opened 2015-01-27
Update wtforms.po (chinese translations)

Aaron1992:
Correct a translation error


davidism:
Thanks, but I'm going with the changes in #165.

PR #159, opened 2015-02-04
Make populate_obj() ignore some fields

ronbeltran:
We're using appengine datastore for our backend. We have a field that is a repeated property

```
emails = ndb.StringProperty(repeated=True) # expects a List of String
```

So `form.populate_obj(my_obj)` fails because in form the input type is string which is comma separated.
We need to have a way to ignore populating some fields so that we can set it with the correct input type.

```
form.populate_obj(my_obj, ignore_fields=['emails',])
my_obj.emails = form.emails.data.split(',')
my_obj.put()
```

### DIFF
--- a/tests/form.py
+++ b/tests/form.py
@@ -67,6 +67,17 @@ class BaseFormTest(TestCase):
         self.assertEqual(m.test, 'foobar')
         self.assertEqual([k for k in dir(m) if not k.startswith('_')], ['test'])
 
+    def test_populate_obj_ignore_fields(self):
+        m = type(str('Model'), (object, ), {})
+        form = self.get_form()
+        self.assertEqual(len(list(form)), 1)
+        form['foo'] = TextField()
+        self.assertEqual(len(list(form)), 2)
+        form.process(DummyPostData(foo=['hello']))
+        form.populate_obj(m, ignore_fields=['test'])
+        self.assertEqual(form['foo'].data, 'hello')
+        self.assertEqual([k for k in dir(m) if not k.startswith('_')], ['foo'])
+
     def test_prefixes(self):
         form = self.get_form(prefix='foo')
         self.assertEqual(form['test'].name, 'foo-test')


prencher:
Happy to accept this if you add unit tests and fix a couple of minor formatting issues: Add empty lines after the if, and after the continue.

ronbeltran:
@prencher Please inform if I missed something :-)

crast:
My only little issue with this is that it's not recursive, it only works with base fields, but otherwise I like this

crast:
Duplicate of #147 

I think we should pull bits from both of these

lafrech:
There seems to be demand for this (https://github.com/wtforms/wtforms/issues/162, https://github.com/wtforms/wtforms/pull/147), so it would make sense to integrate such a feature.

I'd be interested in such a modification as well. Here are a few related reflexions.

### exclude / only

While proposing an `exclude` (or `ignore`) option, why not also add an `only` option that would work as a whitelist (inspired from [MongoEngine](http://docs.mongoengine.org/en/latest/apireference.html#mongoengine.queryset.QuerySet.exclude))?

If both exclude and only are specified, a field would be populated if it is in only and not in exclude.

### exclude disabled fields ###

There is another issue that can be addressed here. The fact that [disabled fields are considered for validation](http://stackoverflow.com/questions/16481924/disabled-field-is-considered-for-validation-in-wtforms-and-flask). This is a noob trap. When setting a field as "disabled", it is rendered as disabled client-side, but nothing prevents a malicious client to inject data in the field, so it is advised to remove the field from the form. But say you display it as disabled, then validation passes, then you remove the disabled field from the form, then, while saving, another error occurs (for instance, a field breaks a NotUnique condition you couldn't easily check at validation time, or any other specific error hard or impossible to catch before saving). What I like to do here is reload edit page showing an error message. But the edit page rendering fails because the field was removed from the form.

The solution I found was to override populate_obj to add an option not to populate disabled fields. It may be addressing a corner case, but I don't think it hurts if False by default.

The overridden method looks like this:

```python
class AutoDisableForm(Form):
    """Form that uses the disabled flag to ignore disabled fields.

       This allows to keep disabled fields for display purpose without
       being open to malicious data injection.
    """

    def populate_obj(self, obj, ignore_disabled=False):
        """
        Populates the attributes of the passed `obj` with data from the form's
        fields, ignoring disabled fields.
        """
        for name, field in iteritems(self._fields):
            if not ignore_disabled and field.flags.disabled:
                field.populate_obj(obj, name)
```



### Recursion

What if a field to be excluded is in a `FormField`? The simplest way to specify this, IMHO, would be to write `exclude=['formfield.field']`.

The only special case I can think of with regard to  `ListField` is if it is a list of `FormField`s. So to exclude a field from a `FormField` in a `Listfield`, we could propagate the same way: `exclude=['listfield.field'.]`.



Before submitting anything, do we agree on the following features?

- new param exclude=[]: blacklist of fields to ignore
- new param only=[]: whitelist (populate only those)
- exclude_disabled=False: exclude disabled fields if set to True
- partial=False: see https://github.com/wtforms/wtforms/pull/147/files
- should be recursive: `exclude` and `only` should work with `FormField` and `FormField` inside `ListField`

kblomqvist:
Any chance to get these merged?

kblomqvist:
Actually I was hitting [this](https://github.com/wtforms/wtforms/blob/086d0bf7d6d73ab9a05e30b861adee627916a71c/wtforms/fields/core.py#L829) (exception from FormField) when trying "partial population". Partial population works without any PRs if I'm not using FormField.

So far my hack is:

```python
from wtforms.fields import FormField as WTFormField

class FormField(WTFormField):

    def populate_obj(self, obj, name):
        candidate = getattr(obj, name, None)
        try:
            self.form.populate_obj(candidate)
        except AttributeError:
            pass
```

davidism:
For the original use case described, this is probably better solved with a custom field that splits during `process_formdata`.

I may revisit this as we explore supporting other data sources, such as patch requests, but for now I'd like to keep `populate_obj` simple. As shown, if you need custom behavior you can override the method.

Abdur-rahmaanJ:
@davidism What is the update to it?

PR #160, opened 2015-02-05
String field fix

volker48:
I noticed that if I have a `StringField` on a form that is being populated by `kwargs`, but I also have some other fields being populated by `formdata` the `StringField` would be set to `''` inside `StringFields` `process_formdata` method even though `self.data` was being set properly in the base `Field` class's `process_data` method. Adds the test `test_formdata_and_kwargs` that fails without my change in `core.py`.

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -5,6 +5,7 @@ import sys
 from datetime import date, datetime
 from decimal import Decimal, ROUND_UP, ROUND_DOWN
 from unittest import TestCase
+from werkzeug.datastructures import MultiDict
 
 from wtforms import validators, widgets, meta
 from wtforms.fields import *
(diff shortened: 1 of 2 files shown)


volker48:
Looks like build failed because pip install failed with `error: [Errno 104] Connection reset by peer` when trying to install SQLAlchemy.

prencher:
This is not a bug: http://wtforms.readthedocs.org/en/latest/faq.html#why-does-blank-input-not-go-back-to-the-default-value.

volker48:
Hmm the reason I made the PR is because I saw inconsistent handling between the Field classes and presumed that there was a bug in the `StringField` class. I noticed this issue because `StringField` and its descendants are the only ones having this issue. A similar test like the one in my pull request, but using an `IntegerField` instead of a `StringField` behaves as I would expect.

```
diff --git a/tests/fields.py b/tests/fields.py
index 7f717f2..57a22d6 100644
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -421,6 +421,10 @@ class IntegerFieldTest(TestCase):
         self.assertEqual(form.a._value(), '')
         self.assertEqual(form.b._value(), '9')
 
+    def test_formdata_and_kwargs(self):
+        form = self.F(DummyPostData({'otherField': 1}), a=42)
+        self.assertEqual(form.a.data, 42)
+
 
 class DecimalFieldTest(TestCase):
     def test(self):
```

The field `a` is populated via the `kwargs`. If you try and do this with a `StringField` you will get the empty string. So either `IntegerField` is incorrect or `StringField` is.

PR #166, opened 2015-02-23
Add max and min values for NumberInput and RangeInput widgets

Informatic:


ftm:
This was done in #233

Issue #117, opened 2014-08-15
validators.Email passes emails with spaces


lexabug:
`validators.Email` passes emails that contain spaces as valid. Please fix the regex to block such emails
```python
>>> from wtforms import validators, Form, fields
>>> from werkzeug.datastructures import ImmutableMultiDict
>>> class MyForm(Form):
...    email = fields.StringField('email', validators=[validators.Email()])
>>> data = ImmutableMultiDict({'email': 'lexabug github@gmail.com'})
>>> form = MyForm(data)
>>> form.validate()
True
>>> form.errors
{}
```

shuisman:
+1

crast:
Too many email validator fix tickets. See comment here: https://github.com/wtforms/wtforms/issues/172#issuecomment-164679909

PR #169, opened 2015-03-03
This code improves email validation

matheusrocha89:
WTforms allowed spaces on email address
this code implements issue #117 solution.

Also do a `trim` on email to now invalidate emails
that accidentaly have spaces at the beggining and the end.

### DIFF
--- a/tests/validators.py
+++ b/tests/validators.py
@@ -23,6 +23,7 @@ class ValidatorsTest(TestCase):
         self.assertEqual(email()(self.form, DummyField('foo@bar.dk')), None)
         self.assertEqual(email()(self.form, DummyField('123@bar.dk')), None)
         self.assertEqual(email()(self.form, DummyField('foo@456.dk')), None)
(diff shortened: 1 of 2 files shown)


matheusrocha89:
@crast , the travis CI of wtforms runs the command `coverage run tests/runtests.py --with-pep8 --force-all` and exit with code one because of some `lambda` that should be replaced by `def`because of pep8 I guess.

crast:
We don't do data trimming on a field in validators, that's inconsistent with how we operate.

The validator should assert the email is valid, not fix the email (this is consistent with all the other validators)

Instead it's better people use the pattern  `StringField(validators=[v.email()], filters=[strip_filter])`  which will allow the filtering to be done in a least-surprise basis;  the data being different before and after to the call to `validate()` does not fit that.

What people should be doing instead of repetetive patterns for fields they use a lot though is something like this:
```python
from wtforms import StringField
from functools import partial
EmailField = partial(StringField, validators=[v.email()], filters=[strip_filter])
```
or alternately subclassing to form their own patterns.

PR #171, opened 2015-03-06
Corrected czech translations, fixed translations propagation in FieldList.

vikipha:


crast:
Great job noticing that FieldList issue.

Forgive me, since I don't personally speak czech, but given you changed almost every translation, can you explain the nature of  Špatně versus Neplatný / Neplatná

vikipha:
Yes, i know it looks weird :-) "Špatně" means "bad" in czech. "Neplatný" is exact translation of word "invalid". In fact message "Not a valid list" was (in the current version) in czech literally "Bad list". Similarly "Not a valid decimal value number" was "Bad decimal number",  "Invalid field name" was "Bad field name"(which is not that bad :-)), "Not a valid date value" was "Bad date" and so on... I think my version is more exact and consistent :-)

PR #173, opened 2015-03-10
Use polyglot markup

yggi49:
Polyglot markup has the advantage of validating as both HTML5 and XML 1.0 at the same time.  Therefore, a web page using polyglot markup could, e.g., either be served by the `text/html` or the `application/xhtml+xml` MIME type, or be further processed by XML tools—depending on the project’s application-specific requirements.

This pull request adds support for polyglot markup by correctly closing `input` elements, and providing a value for boolean HTML attributes.  Users who generate “standard” HTML pages won’t notice a difference.

For further information, please refer to the [Introduction][1] and [Principles][2] sections of the W3C candidate recommendation [Polyglot Markup: A robust profile of the HTML5 vocabulary][3].

*Edit: see [Pull Request #174][4] for a different approach.*

[1]: http://www.w3.org/TR/html-polyglot/#introduction
[2]: http://www.w3.org/TR/html-polyglot/#principles
[3]: http://www.w3.org/TR/html-polyglot/
[4]: https://github.com/wtforms/wtforms/pull/174

PR #174, opened 2015-03-11
Use polyglot markup—a different approach

yggi49:
This is a different approach to introduce polyglot markup for widget rendering.

In contrast to [Pull Request #173][1], which unconditionally modifies all output, this change introduces a new base form class, `PolyglotForm`.  It utilizies a new custom form meta class, `PolyglotMeta`, which transforms the rendered widgets’ output to be polyglot-compatible.  This way, users can deliberately opt-in to using polyglot markup for their HTML output.

For further information, please refer to the [Introduction][2] and [Principles][3] sections of the W3C candidate recommendation [Polyglot Markup: A robust profile of the HTML5 vocabulary][4].

[1]: https://github.com/wtforms/wtforms/pull/173
[2]: http://www.w3.org/TR/html-polyglot/#introduction
[3]: http://www.w3.org/TR/html-polyglot/#principles
[4]: http://www.w3.org/TR/html-polyglot/

### DIFF
--- a/tests/polyglot.py
+++ b/tests/polyglot.py
@@ -1,0 +1,121 @@
+from unittest import TestCase
+
+from wtforms import (PolyglotForm, StringField, TextAreaField, BooleanField,
+                     SelectField, SelectMultipleField)
+from wtforms.meta import PolyglotHTMLParser
+
+
+class PolyglotHTMLParserTest(TestCase):
+
+    def setUp(self):
+        self.parser = PolyglotHTMLParser()
+
+    def test(self):
+        self.parser.feed('<input type=checkbox name=foo value=y checked>')
+        self.assertEqual(self.parser.get_output(),
+                         '<input type="checkbox" name="foo" value="y"'
+                         ' checked="checked" />')
+
+    def test_bad_standalone_tag(self):
+        self.parser.feed('<input>')
+        self.assertEqual(self.parser.get_output(), '<input />')
+
+    def test_good_standalone_tag(self):
+        self.parser.feed('<input />')
+        self.assertEqual(self.parser.get_output(), '<input />')
+
+    def test_bad_standalone_tag_with_unquoted_attribute(self):
+        self.parser.feed('<input name=foo>')
+        self.assertEqual(self.parser.get_output(), '<input name="foo" />')
+
+    def test_good_standalone_tag_with_unquoted_attribute(self):
+        self.parser.feed('<input name=foo />')
+        self.assertEqual(self.parser.get_output(), '<input name="foo" />')
+
+    def test_bad_standalone_tag_with_quoted_attribute(self):
+        self.parser.feed('<input name="foo">')
+        self.assertEqual(self.parser.get_output(), '<input name="foo" />')
+
+    def test_good_standalone_tag_with_quoted_attribute(self):
+        self.parser.feed('<input name="foo" />')
+        self.assertEqual(self.parser.get_output(), '<input name="foo" />')
+
+    def test_bad_standalone_tag_with_bad_boolean_attribute(self):
+        self.parser.feed('<input checked>')
+        self.assertEqual(self.parser.get_output(),
+                         '<input checked="checked" />')
+
+    def test_good_standalone_tag_with_bad_boolean_attribute(self):
+        self.parser.feed('<input checked />')
+        self.assertEqual(self.parser.get_output(),
+                         '<input checked="checked" />')
+
+    def test_non_standalone_tag(self):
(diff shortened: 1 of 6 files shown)


prencher:
Not a bad approach, but I do not think it belongs in core. Instead I think it would be a good candidate for a wtforms-xhtml or wtforms-polyglot companion library, which can just depend on wtforms, and the only change would have to be `from wtforms import Form` becomes `from wtforms_polyglot import Form`.

yggi49:
prencher, thank you for reviewing my pull requests.  In fact, I strongly believe that changing WTForms’ default output would be the way to go, as in #173.  Most users do not care about the markup of their web pages anyway.  For those who do, polyglot markup is an excellent choice, as it does not arbitrarily limit them in how they use the library.  In today’s web environments, the additional cost of producing a few—and we are really talking only a few—bytes of HTML more is negligible in my opinion.

Be it as it may, I will create a companion library as you suggested—both for WTForms and also for Flask-WTF, which is my main use case for WTForms.

prencher:
We actually changed it primarily due to requests to do so, and with HTML 5 rapidly becoming the norm at the time. I have no preference since it's all just generated code, but people needing XHTML at this point are in a very significant minority, so changing it back would break compatibility for very little gain.

The parsing solution is fine, but it is not suitable for a core feature; I'd be concerned it potentially be too fragile for custom fields without using a full on HTML parser, which we wouldn't want to depend on.

I'd suggest that you simply make flask an optional dependency and create a flask module inside to extend the flask forms.

We're thinking about WTForms 3.0, and have some ideas to make it easier to affect all widgets, as opposed to overriding all fields.

yggi49:
I share your opinion that the parsing solution is potentially fragile.  This is why I prefer #173. ;-)  Apart from that, I see your point, but I am unable to grasp how using polyglot markup would break any compatibility.  

For clarification: polyglot markup *is* 100 % valid HTML5; in fact, it is a subset of HTML5.  (Imagine it as being just a ruleset on how to write HTML5.)  For example, HTML5 explicitly allows using the XML-ish slash in front of a closing angle bracket of void elements like `input`.  Both the [W3C spec][w3c-start-tags] and the [WHATWG spec][whatwg-start-tags] have in their “Start tags” subsection of the “Syntax” section (item 6 in each case):

> Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

*Note: “foreign elements” refers to MathML- and SVG-related elements, while “void elements” stands for all of the ‘usual’ HTML elements that have no children, like `input`, `img`, etc.*

Likewise, HTML5 explicitly allows boolean attributes to use the attribute’s name as its value.  Again, this is present in the “Boolean Attributes” section of both the [W3C spec][w3c-boolean-attributes] and the [WHATWG spec][whatwg-boolean-attributes]:

> If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.

Hence, all of the following are perfectly valid HTML5 snippets, without even thinking of XML:

```html
<input type=checkbox name=foo value=y checked>
<input type=checkbox name=foo value=y checked=checked>
<input type="checkbox" name="foo" value="y" checked>
<input type="checkbox" name="foo" value="y" checked="checked">
<input type=checkbox name=foo value=y checked />
<input type=checkbox name=foo value=y checked=checked />
<input type="checkbox" name="foo" value="y" checked />
<input type="checkbox" name="foo" value="y" checked="checked" />
```

It just so happens that if you adhere to the polyglot ruleset for writing your HTML5, your document coincidentally also validates as XML. ;-)

(Sorry for the long text, but I really care a lot.)

[w3c-start-tags]: http://www.w3.org/TR/html5/syntax.html#start-tags
[whatwg-start-tags]: https://html.spec.whatwg.org/multipage/syntax.html#start-tags
[w3c-boolean-attributes]: http://www.w3.org/TR/html5/infrastructure.html#boolean-attributes
[whatwg-boolean-attributes]: https://html.spec.whatwg.org/multipage/infrastructure.html#boolean-attributes

prencher:
Not incompatibility with browsers, but a breaking change in WTForms. Anybody testing their output would need to update tests, as would we for the core tests.

That requires a major version bump and I do not think it is worth it just for this.

yggi49:
Point made, thanks for clarification!

yggi49:
[`WTForms-Polyglot`][1] is now available on PyPI, providing `PolyglotForm`.

[1]: https://pypi.python.org/pypi/WTForms-Polyglot

PR #178, opened 2015-03-14
Allow obj to be dict or object

wspeirs:
For folks using document databases to persist forms, it would be really helpful to allow obj in Form construction to be either an object (as it currently is) or a dict (which is commonly read back from document DBs such as Mongo or Rethink).

This pull request makes a simple change to check the type of obj and if it's a dict, then uses the value found at the name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ dist/
 docs/_build
 docs/html
 env/
+
+# IDE
+.idea*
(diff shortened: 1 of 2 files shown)


wspeirs:
Curious if anyone had a chance to take a look at this enhancement... thanks!

crast:
You can pass dictionaries as `data=` to the form constructor, or by expanding out keyword args.

PR #196, opened 2015-06-11
Catch TypeError in IntegerField

Kazanz:
`wtforms.fields.core.IntegerField.process_formdata` only catches `ValueErrors` when coercing to int. 

If a non string is passed to the method it will throw a `TypeError`.  If the user can pass an invalid string and have it caught in an try-except block, they should also be able to pass other invalid types as well.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -579,7 +579,7 @@ class IntegerField(Field):
         if valuelist:
             try:
                 self.data = int(valuelist[0])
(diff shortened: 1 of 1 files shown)


crast:
This can't happen if WTForms is used with an appropriate input wrapper

I'm aware that there are some other frameworks which do evil things to try to represent JSON to WTForms somehow, and wtforms 2.x doesn't support JSON input

PR #200, opened 2015-06-29
Fix DataRequired validator

lepture:
Consider IntegerField. If the value is 0. In this case, the field has a value, but
DataRequired with `not field.data` will raise an Exception.


crast:
Use InputRequired, not DataRequired. DataRequired is there for legacy compatibility reasons

PR #209, opened 2015-07-20
Coerce choices on validation

jmagnusson:
This feature was created to be able to avoid using the string representation of the choice in the `option` element's `value` attribute. E.g:

``` python
CHOICES = (
    (None, 'Yes/No'),
    (True, 'Yes'),
    (False, 'No'),
)
```

would turn into `<option value="None">`, `<option value="True">` and `<option value="False">` before.

With this commit we can now use:

``` python
CHOICES = (
    ('', 'Yes/No'),
    ('yes', 'Yes'),
    ('no', 'No'),
)
```

If we make sure to pass a custom `coerce` function that turns them into "python values":

``` python
def yes_no_none_coercer(val):
    try:
        return {'': None, 'yes': True, 'no': False}[val]
    except KeyError:
        raise ValueError(val)

        def yes_no_none_coercer(val):
            try:
                return {'yes': True, 'no': False, '': None}[val]
            except KeyError:
                raise ValueError(val)

class ProductDetails(Form):
    has_image = SelectField('Has image', choices=CHOICES,
                            default='', coerce=yes_no_none_coercer)

form = ProductDetails(has_image=['yes'])
assert form.validate()
assert form.has_image.data == True
assert 'value="yes"' in str(form.has_image)
```

The `coerce=int` examples in the documentation and tests still works.

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -17,6 +17,20 @@ from tests.common import DummyPostData
 PYTHON_VERSION = sys.version_info
 
 
+def yes_no_coercer(val):
+    try:
+        return {'yes': True, 'no': False}[val]
+    except KeyError:
+        raise ValueError(val)
+
+
+def yes_no_none_coercer(val):
+    try:
+        return {'': None, 'yes': True, 'no': False}[val]
+    except KeyError:
+        raise ValueError(val)
+
+
 class AttrDict(object):
     def __init__(self, *args, **kw):
         self.__dict__.update(*args, **kw)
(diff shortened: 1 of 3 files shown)


tests/fields.py
```diff
@@ -296,6 +336,13 @@ def test_coerce_fail(self):
         assert not form.validate()
```
jmagnusson: Why does this validate? I would love to hear an explanation.

tests/fields.py
```diff
@@ -288,6 +315,19 @@ def test_with_data(self):
...
+            self.assertEqual(form.c.data, vals)
+            self.assertTrue(form.validate())
+            for input_val in input_vals:
+                self.assertTrue('value="%s"' % input_val in str(form.c))
+            for val in vals:
+                self.assertFalse('value="%s"' % val in str(form.c))
+
     def test_coerce_fail(self):
         form = self.F(b=['a'])
         assert form.validate()
```
jmagnusson: Why does this validate? Not sure I understand why...

ftm:
I don't really see what the point of this would be. When creating the choices you already specify value/internal value and the user facing value, why bother coercing them to another value when you could've just used that value in the first place?

PR #213, opened 2015-08-10
Render the form object as a trimmed down dictionary

jamesboehmer:
Added a dict property which transforms the form into a dictionary of inputs (which retain the _fields order), and errors.  The output of the dict property can be easily serialized to JSON or XML.  The idea is to be able to use the same WTForms objects in your markup as you do in your APIs.

e.g.

``` python
new_form = SomeWTForm()
    if new_form.is_submitted():
        if new_form.validate():
            # do some stuff
        else:
            return Response(json.dumps(new_form.dict), 400)
    else:
        return Response(json.dumps(new_form.dict), 200)
```


crast:
What's wrong with `form.data` ?

davidism:
This seems more appropriate for an extension.

PR #216, opened 2015-09-03
incorrect polish translations

squfrans:
these translations are not correct.
unfortunately i can only read a bit of polish,
but i am not a fluent speaker, and i don't have
any developer friends speaking polish.

i think removing these translations is better
than to have the misleading ones.


squfrans:
ah, sorry, i just saw #206, closing this one

PR #219, opened 2015-09-15
Allow widgets to set flags and make HiddenInput set the "hidden" flag.

anton-d:
It is sometimes useful to be able to determine in a template whether a field is hidden (e.g. to not display the label). To this end the present change lets the HiddenInput widget set the "hidden" flag on the parent field. More generally this change lets all widgets set flags, in the same way as validators.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -109,6 +109,9 @@ class Field(object):
 
         if widget is not None:
             self.widget = widget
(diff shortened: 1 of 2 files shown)


crast:
That's actually... really dang smart. And obvious and simple.

Thanks for the contribution, just needs some test coverage before I accept it, so if you can add tests it'll get accepted quicker

PR #224, opened 2015-10-21
Optgroup

hsum:
Goals of this PR:
- allow specifying of attributes for options within a select at init time
- allow specifying optgroups for options at init time.  This uses the recently added (Field) render_kw argument.

Motivation:
- allow better select widgets (e.g. integrated with Select2)

Details:
- SelectField.choices can accept an iterable or a callable that returns/yields an iterable.
- choices are of class OptionChoice, which is a subclass of a namedtuple.  Choices can still be an iterable of tuple(value, label) .  Remaining backward compatible is important because (value, label) is usually sufficient.
- optgroup is specified by providing a group key function (same as itertools.groupby).  The label for the optgroup is also specified as a function that received some context variables (e.g. the group of OptionChoice objects, enumeration and the key used to create this group).  If the label is not specified, the label will default to the enumeration of this group (starting at 0 because enumerate() does so).
- tests for the above pass
- extensions need testing, sqlalchemy seems to work for me

Please let me know what you think.

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -228,6 +228,47 @@ class SelectFieldTest(TestCase):
     class F(Form):
         a = SelectField(choices=[('a', 'hello'), ('btest', 'bye')], default='a')
         b = SelectField(choices=[(1, 'Item 1'), (2, 'Item 2')], coerce=int, option_widget=widgets.TextInput())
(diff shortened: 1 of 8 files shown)


ftm:
This is quite a large PR with a lot of merge conflicts. Whilst it is a very good idea (one which I can see us coming back to) I feel like in its current form there's too much work to do so I'm going to close it for now.

PR #229, opened 2015-11-10
fix SelectField.pre_validate

graycarl:
Hi, I think you miss something in `SelectField`. The value from `self.choices` should be `coerce` before compare to `self.data`.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -466,7 +466,7 @@ class SelectField(SelectFieldBase):
 
     def pre_validate(self, form):
         for v, _ in self.choices:
-            if self.data == v:
+            if self.data == self.coerce(v):
                 break
         else:
             raise ValueError(self.gettext('Not a valid choice'))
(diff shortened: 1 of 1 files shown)


crast:
it's not input data, you can coerce it yourself

Won't be fixed in 2.x (is a breaking change) and won't likely be relevant in 3.x

graycarl:
I just want to explain my point.

When you `iter_choices`:

```
    def iter_choices(self):
        for value, label in self.choices:
            yield (value, label, self.coerce(value) == self.data)
```

you coerce the value before compare to `self.data`, but in `pre_validate` not. That make things inconsistent.

dkasak:
@crast Could removing the coercion in `iter_choices` at least be considered for 2.x since you yourself say that the reference values in `choices` shouldn't get coerced? [This](https://github.com/dkasak/wtforms/commit/bf8a282c5628d9e91d0a014118a765166a12ea96) is all that needs to be done to fix this and I don't see how it is a breaking change.

PR #231, opened 2015-11-17
Fix "KeyError: u'max'" On Length Validator

pawl:
I'm using flask-babelex for translations. When I try to raise a Length ValidationError, I get a KeyError because the message isn't using the default "%(num)d" or a dict. This pull request changes "%(min)d" and "%(max)d" in the translations to "%(num)d".

Here's the whole gist if you're interested: https://gist.github.com/anonymous/a6210e203faa1a6c8ebc


pawl:
Just realized I had to update translations to fix this. Updated the pull request.

pawl:
@crast can you take a look at this? The change is much smaller than it looks, mainly just changing "min" to "num" here so it jives with ngettext defaults: https://github.com/wtforms/wtforms/pull/231/files#diff-f98d8f668897e84e001195c93ac781a7R97

whb07:
I'll have to

PR #241, opened 2016-01-04
Update forms.rst

halcyonjuly7:
Lacking self in inline validators
`line 142`

### DIFF
--- a/docs/forms.rst
+++ b/docs/forms.rst
@@ -139,7 +139,7 @@ method with the convention `validate_fieldname`::
     class SignupForm(Form):
         age = IntegerField(u'Age')
 
-        def validate_age(form, field):
+        def validate_age(self, form, field):
             if field.data < 13:
                 raise ValidationError("We're sorry, you must be 13 or older to register")
 


tmehlinger:
I noticed this myself recently. The docs are not actually missing a parameter, rather, I suspect this is a wtforms convention where `form` is given in place of `self`. I would argue that's a bad convention, but we'll see where the discussion goes.

This code will fail with `TypeError: validate_domain() missing 1 required positional argument: 'field'` when called. Here's the relevant code:

https://github.com/wtforms/wtforms/blob/master/wtforms/form.py#L306

https://github.com/wtforms/wtforms/blob/master/wtforms/fields/core.py#L224

From the first link, you can see how the form sets up the inline validators. It uses `getattr` to make a list of _unbound_ methods from the form class. From the second link, note the two parameters, `form`, and `self`, where `form` is the instance of the form that defines the field being validated, and `self` is the field itself. I suspect the convention from the docs reflects the behavior of the code--passing a form instance as the first parameter to an unbound method in a form class.

I'd like to see the docs changed as follows to adhere to Python convention:
```
class SignupForm(Form):
    age = IntegerField(u'Age')
 
    def validate_age(self, field):
        if field.data < 13:
            raise ValidationError("We're sorry, you must be 13 or older to register")
```

crast:
@tmehlinger I think that's a good idea. Will accept that patch for the docs; or I may get to it sooner

PR #247, opened 2016-01-28
Bugfix: propagate meta instance to form in FormField

jeinarsson:
The child form in a `FormField` does not have access to the `meta` instance of the parent form. Unless this is intended behavior I made a bugfix with three changes:
- `Form` constructor argument `meta` now accepts a meta object as alternative to a `dict`.
- `FormField` propagates `self.meta` to child form via new constructor logic.
- Regression tests.

This patch works also on 2.1 (that I use in my application, manually patched)

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -131,6 +131,7 @@ class FiltersTest(TestCase):
         self.assertEqual(len(form.b.process_errors), 1)
         assert not form.validate()
 
+
 class FieldTest(TestCase):
     class F(Form):
         a = TextField(default='hello', render_kw={'readonly': True, 'foo': u'bar'})
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -131,10 +131,10 @@ class FiltersTest(TestCase):
         self.assertEqual(len(form.b.process_errors), 1)
         assert not form.validate()
 
-
 class FieldTest(TestCase):
     class F(Form):
         a = TextField(default='hello', render_kw={'readonly': True, 'foo': u'bar'})
+        formfield = FormField(Form)
 
     def setUp(self):
         self.field = self.F().a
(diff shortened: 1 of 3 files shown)


jeinarsson:
If someone else on wtforms 2.1 bumps into this, I made an unofficial patched 2.1.1 branch in my fork. You can install it by

`pip install git+https://github.com/jeinarsson/wtforms.git@fix-formfield-meta#egg=WTForms `

crast:
One potential issue here is that if you are using the CSRF feature, it would then enable CSRF on the child form, but the CSRF would never validate due to the field not being sent with that prefix. I think we'd probably need tests to that effect

jeinarsson:
Good point.

Option one: make CSRF on child forms work seamlessly (except one has to render the fields). Option two: make child forms not have CSRF.

In my opinion child forms do not need CSRF, so I prefer option two. A custom CSRF implementation should not be bothered with this detail, therefore this must be implemented in core.

I see two straightforward ways:
A) Give the child form a meta instance that is a copy of parent meta but csrf = False. This is slightly hacky and is perhaps unexpected behavior if a user changes a value in the parent meta and expects child to be updated.
B) Add a flag to Form constructor signalling whether the Form is a child, and if it is don't construct any CSRF, eg a 'is_child' kwarg. FormField passes this flag when creating child Form instances. 

I prefer option B, because the conflict is a direct consequence of placing the csrf config in meta, and at the same time allowing FormFields. Therefore flagging child forms for special treatment seems like an OK resolution.

Did I miss anything obvious? Cheers

ftm:
What about a way to pass in the "parent" form in the constructor. Then you don't have to deal with unintended things happening to the child's Meta class (e.g. CSRF) and you can pick and choose which parts of the parent's Meta class you want in the child.

Somewhere you could have a check of if `self.parent is None` then disable CSRF

ftm:
Due to the merge conflicts caused by recent changes to the repo and the age of this pull request I'm going to close it

PR #248, opened 2016-01-28
hyphenate all data-attribute underscores

doobeh:
data-x-y is a much more commonly used naming scheme than data-x_y.  Popular frameworks such as [Bootstrap contains data-x-y attributes](http://getbootstrap.com/javascript/#buttons-methods) as does [Foundation](http://foundation.zurb.com/sites/docs/javascript.html) and [uikit](http://getuikit.com/docs/lightbox.html).  

As mentioned in the [issue comments](https://github.com/wtforms/wtforms/issues/48)— the spec only mentions `data-*` but its clear from just hitting Google with `data- attributes html` that all the top examples show multiple hyphenated usage, so in this case I think there's a greater win in building towards the common convention rather then the spec as stated.

### DIFF
--- a/tests/widgets.py
+++ b/tests/widgets.py
@@ -32,7 +32,7 @@ class HTMLParamsTest(TestCase):
 
     def test_data_prefix(self):
         self.assertEqual(html_params(data_foo=22), 'data-foo="22"')
-        self.assertEqual(html_params(data_foo_bar=1), 'data-foo_bar="1"')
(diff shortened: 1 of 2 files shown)


crast:
The main issue was WTForms 2.0 compatibility promise. In other words, not breaking behaviours for the sake of breaking behaviours.

But now that we're on 3.0 preview, I can merge this.

PR #253, opened 2016-02-08
Fix typo in _run_validation_chain documentation

djds23:


djds23:
Not sure why coveralls is broken. Only corrected a single documentation line.

djds23:
thanks @crast!

PR #258, opened 2016-02-20
fields: keep entries order in errors list

the-tosh:


wtforms/fields/core.py
```diff
@@ -915,16 +915,23 @@ def validate(self, form, extra_validators=tuple()):
...
+                self.errors.append(None)
+
+        if all(x is None for x in self.errors):
+            self.errors = []
 
         chain = itertools.chain(self.validators, extra_validators)
         self._run_validation_chain(form, chain)
 
-        return len(self.errors) == 0
+        return not error_occured
```
valtron: This ignores errors from `_run_validation_chain` (hence the failing test). What about `return self.errors and None not in self.errors`?

the-tosh: Hmmm. Yes, it will work, I guess.
Should I add the changes to this pull request? I'm still bad with pull requests, so I'm not sure it's even possible in technical meaning :)

valtron: Yeah, you can add more changes, it's as simple as pushing to `the-tosh:master` 👍

the-tosh: Done. Thanks!

valtron: Sorry about this, but my initial comment had a wrong suggestion :( The line should be `return all(e is None for e in self.errors)`.

the-tosh: The funniest thing here, that I've been already about typing `all`... and then I just copypaste line from your message :)  Well, fixed.

whb07:
Thanks for trying to contribute! Having said that this PR isn't refining what it appears to be doing. Adding `None` to the errors list and then having to do more work to check for the thing we just added is unnecessary. 

As of now I'm going to be closing the PR. If you happen to find anything that you'd like to help out with dont hesitate to submit another PR.

davidism:
To clarify, I think what we're saying is that the issue can be solved by rendering the errors for the subfields individually. Perhaps you're using a template macro and it can be adjusted to understand nested fields.

valtron:
@whb07 Looking at `FieldList.errors` (or `.errors` of any field enclosure containing it), it's impossible to determine which list item has errors. Code:

```python
from multidict import MultiDict
from wtforms import FieldList, validators, Form, TextField

class TestForm(Form):
	test_field = FieldList(TextField("Test", [validators.required()]))

data = MultiDict([('test_field-{}'.format(i), value) for i, value in enumerate(['', 'a'])])
form = TestForm(data)
assert not form.validate()

print(form.errors)
# This would imply list item 0 has an error:
# {'test_field': [['This field is required.']]}

for f in form.test_field: print(f.errors)
# ['This field is required.']
# []
```

@davidism It's possible to work around it, like you say. But the problem then becomes that `FieldList.errors` is only usable for detecting and counting errors *somewhere* in the list, but not identifying where they are. (Contrast with `FormField`, where there's no problem because subfields are keyed.)

My use case is AJAX form validation. I respond with `form.errors` and zip down it client-side to show errors in the right place.

whb07:
Okay this newest message shows me what you're actually trying to say. I do think that this issue of the FieldList not keeping track of the data passed to it is something that should be worked on. Having said that, the current PR as it stands is not doing that.

So clearly when you pass in the data such as `test_field-0-name_of_field` the form is correctly associating the data to the field. But you want it to return which specific field it was correct?

If thats your goal, I'd like to help out. Open a new PR with possible changes and tests verifying what you are talking about. I'll be happy to take a look at it as I've also run into this issue.

PR #262, opened 2016-03-09
Hungarian translation added

zoltan-fedor:
Added Hungarian translation


davidism:
I'm aware of this as well as the other Hungarian translation in #268 .  I asked a friend who speaks Hungarian to look over both, they said:

> the two translations are almost equally valid, and both are lacking a few commas in my opinion. Anyway, I'm leaning a bit towards 262, it's a bit more specific in a few places. But watch out for line 99: the field specifier is missing from the end of the translation. And "not a valid float" is translated into "not a valid fraction", which is bugging me a bit (line 169 in the other translation is precise)

Since this PR came first and it has an external vote of approval I'm more inclined to merge this one assuming the two things mentioned get fixed.  Could you give this another look over and compare notes with #268?

zoltan-fedor:
Hi @davidism ,
I have fixed the two issues mentioned after comparing to #268

Issue #261, opened 2016-03-02
ngettext fails in validators


piranha:
So I have an app using flask-admin and it fails for me with an error:

```
  File "/v1/mkadmin/env_mkadmin/lib/python2.6/site-packages/wtforms/validators.py", line 101, in __call__
    'Field cannot be longer than %(max)d characters.', self.max)
  File "/v1/mkadmin/env_mkadmin/lib/python2.6/site-packages/wtforms/fields/core.py", line 174, in ngettext
    return self._translations.ngettext(singular, plural, n)
  File "/v1/mkadmin/env_mkadmin/lib/python2.6/site-packages/flask_admin/babel.py", line 21, in ngettext
    return ngettext(singular, plural, n)
  File "/v1/mkadmin/env_mkadmin/lib/python2.6/site-packages/flask_admin/babel.py", line 10, in ngettext
    return (singular if num == 1 else plural) % variables
KeyError: u'max'
```

So it seems that you have to pass `{'max': self.max}` to ngettext to get it working.

citizen-stig:
I've noticed the same issue, I use python 3.5

citizen-stig:
Steps to reproduce.

Install flask-admin without Babel, so Flask-Admin will use their own gettext implementation. This gettext tries to interpolate string, but without appropriate variables.

mrjoes:
This was indeed problem in Flask-Admin and it was fixed.

piranha:
Cool, thanks everybody!

dxmo:
Hey,
I'm having similar issue although I am using flask-babelex. Any idea what the solution/workaround would be possible here?

Thanks for any tips!

`Traceback (most recent call last):
  File "/Users/igor/Documents/Tablehead/TableheadCMS/ENV/lib/python2.7/site-packages/flask/app.py", line 1994, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/igor/Documents/Tablehead/TableheadCMS/ENV/lib/python2.7/site-packages/flask/app.py", line 1985, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/igor/Documents/Tablehead/TableheadCMS/ENV/lib/python2.7/site-packages/flask/app.py", line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/igor/Documents/Tablehead/TableheadCMS/ENV/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
[...]

PR #263, opened 2016-03-09
Fix number range error message generation

citizen-stig:
Pull request for #261 .


ftm:
The corresponding issue was closed due to the problem apparently being part of a different project so I'm closing this.

PR #265, opened 2016-03-21
Allow Windows-style MAC address in validator MacAddress

philippefaes:
Users often find their MAC address on a Windows computer. Windows uses dashes "-" instead of colons ":" to separate octets. This patch allows for Windows-style MAC addresses.

### DIFF
--- a/tests/validators.py
+++ b/tests/validators.py
@@ -60,6 +60,7 @@ class ValidatorsTest(TestCase):
         self.assertRaises(ValueError, ip_address, ipv4=False, ipv6=False)
 
     def test_mac_address(self):
+        # Unix-style MAC address
         self.assertEqual(mac_address()(self.form,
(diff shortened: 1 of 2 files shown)


mdxs:
Unfortunately, at the moment your proposed RegEx `^(?:[0-9a-fA-F]{2}(:|-)){5}[0-9a-fA-F]{2}$` solution also validates "01-23:45-67:ab-CD", which is clearly incorrect.

mdxs:
Changing the proposed RegEx to `^((?:[0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2}|(?:[0-9a-fA-F]{2}-){5}[0-9a-fA-F]{2})$` could fix this, as it allows one or the other notation but no mix.

davidism:
I dislike complicated regex. If you need accurate validation in more cases, use the ipaddr library.

PR #266, opened 2016-03-21
Better notification when min and max string length are the same

philippefaes:
Sometimes a string of a fixed length is expect in a form. If you use the `Length` validator, and choose the same value for `max` and `min`, this patch makes the error message more sensible: "Field must be exactly 'n' characters long."

### DIFF
--- a/tests/validators.py
+++ b/tests/validators.py
@@ -98,6 +98,7 @@ class ValidatorsTest(TestCase):
         self.assertEqual(length(min=6)(self.form, field), None)
         self.assertRaises(ValidationError, length(max=5), self.form, field)
         self.assertEqual(length(max=6)(self.form, field), None)
+        self.assertEqual(length(min=6,max=6)(self.form, field), None)
 
         self.assertRaises(AssertionError, length)
         self.assertRaises(AssertionError, length, min=5, max=2)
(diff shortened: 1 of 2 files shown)


whb07:
Hey Phillip I just ran a series of tests setting the Length validator to have a matching min/max values and I wasn't able to reproduce what you're talking about. If you are able to properly document such an event and come back with what you found I'd be more than happy to look at your work further! 

If you're trying to get some sample data for the tests checkout Hypothesis for property based testing. I also used Faker on the side for random text.

ftm:
@whb07 I'm not entirely sure why you've closed this, @philippefaes was submitting a working patch that I was just working on merging

whb07:
whoops! I haven't been able to recreate the error my bad! I ran a series of tests locally and I've yet to see anything that triggers it. Mind showing me a case where it breaks?

ftm:
@whb07 Nothing breaks, this is just making the error message more helpful. Previously it might say something like _Field must be between 5 and 5 characters long_ whereas now it will say _Field must be exactly 5 characters long_

whb07:
oh yah! My bad I didn't review the message properly and thought they somehow ran into an issue with min/max not validating properly. nice patch 👍

PR #272, opened 2016-04-24
Allow labels to be expressed as placeholders

toddjcrane:
Signed-off-by: Todd Crane todd@toddcrane.com


davidism:
A placeholder can be set with `render_kw`: `StringField(render_kw={'placeholder': 'Name'})`. What you want can be done in general by overriding `Meta.render_field`. Since placeholders are not a replacement for labels in terms of accessibility, I don't want to merge special behavior in here.

```python
class PlaceholderBase(Form):
    class Meta:
            render_kw.setdefault('placeholder', field.label.text)
            return super().render_field(field, render_kw)
```

PR #273, opened 2016-05-13
Fix form subclassing.

tvuotila:
Fix form initialization when form has mixins.

I tried to subclass Form with mixin and mixin's `__init__` was never called. This PR fixes the issue.
Example of how I use the mixin:

```
from __future__ import print_function
from wtforms import Form, DateTimeField, TextField, TextAreaField, IntegerField, validators

class ProfileForm(Form):
    birthday  = DateTimeField('Your Birthday', format='%m/%d/%y')
    signature = TextAreaField('Forum Signature')

class AdminMixing(object):
    username = TextField('Username', [validators.Length(max=40)])
    level    = IntegerField('User Level', [validators.NumberRange(min=0, max=10)])

    def __init__(self, *args, **kwargs):
        super(AdminMixing, self).__init__()
        print('called')
        # Some initialization code

class AdminProfileForm(ProfileForm, AdminMixing):
    pass

form = AdminProfileForm()
```

Print shows that `__init__` is not called in the master branch but it is called in this branch.

### DIFF
--- a/tests/form.py
+++ b/tests/form.py
@@ -144,6 +144,14 @@ class FormTest(TestCase):
             if field.data != 'foobar':
                 raise ValidationError('error')
 
+    class P(object):
+        def __init__(self, *args, **kwargs):
+            self.custom_value = True
+            super(FormTest.P, self).__init__()
+
+    class C(Form, P):
+        pass
+
     def test_validate(self):
         form = self.F(test='foobar')
         self.assertEqual(form.validate(), True)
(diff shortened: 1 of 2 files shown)


tvuotila:
Is there something wrong with the Pull Request?

whb07:
As it stands this is an issue with the MRO and not particular to WTForm's BaseForm class. If you're trying to be able to use mixins and define your own classes please be aware of the rules for MRO.

For context:
[SO answer for new style classes MRO ](https://stackoverflow.com/questions/1848474/method-resolution-order-mro-in-new-style-classes)

[Python MRO](https://www.python.org/download/releases/2.3/mro/)

PR #280, opened 2016-06-22
field.process_formdata should still be called if formdata is empty

davidism:
Some fields will set a default value if their key is not in `formdata`.  However, `process_formdata` is only called if `formdata` is truthy.  This can cause unexpected field data if `formdata` is empty versus if it just didn't contain the field.

~~~python
class F(Form):
    name = StringField()

f = F(MultiDict())
print(f.name.data)  # None

f = F(MultiDict({'language': 'Python'}))
print(f.name.data)  # ''
~~~

This patch explicitly checks if `formdata` is `None` and adds a test for this case.

### DIFF
--- a/tests/form.py
+++ b/tests/form.py
@@ -203,6 +203,12 @@ class FormTest(TestCase):
         form = self.F(data=data, test='bar')
         self.assertEqual(form.test.data, 'bar')
 
+    def test_empty_formdata(self):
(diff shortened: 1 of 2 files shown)


hyperknot:
Is there no case when the block of code would handle an exception? I mean it was moved out of a try-except block, and for example if `formdata.getlist(self.name)` produces an exception it is now not handled.

PR #281, opened 2016-06-22
Add support for multiple file input

davidism:
Took a similar approach to `SelectField` and `SelectMultipleField`.  The `FileInput` widget now takes a `multiple` argument, and there are two fields `FileField` and `MultipleFileField`.

`FileField` is no longer based on `StringField`, since it's only a coincidence that the data is a string. I updated the docs to clarify that extensions might change what data is passed to this field (see Flask-WTF for example).

If no data is passed to the field, its `data` is `None` rather than the empty string.  There was no test for this before, so I figured it was ok to break this.  `None` makes more sense for "no file was uploaded".  The multiple field will have an empty list for no data.

`FileInput` always sets `value` to `False` so that it is not rendered.  Browsers won't honor the value of a file input for security reasons anyway.  `FileField._value` returns `False` as well, to be consistent if the widget it changed.

An alternate implementation could use one `FileField` with a `multiple` argument as well.  I'm willing to implement that instead if it sounds better.

This patch contains #280 (call process_formdata if formdata is empty) since testing this ran into that issue.  Merging this after that, or just closing that, should both work in git.

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -16,7 +16,7 @@ Fields are defined as members on a form in a declarative fashion::
         address = TextAreaField(u'Mailing Address', [validators.optional(), validators.length(max=200)])
 
 When a field is defined on a form, the construction parameters are saved until
-the form is instantiated. At form instantiation time, a copy of the field is 
+the form is instantiated. At form instantiation time, a copy of the field is
 made with all the parameters specified in the definition. Each instance of the
 field keeps its own field data and errors list.
 
@@ -63,7 +63,7 @@ The Field base class
 
     .. automethod:: process(formdata [, data])
     .. automethod:: process_data
-    .. automethod:: process_formdata 
+    .. automethod:: process_formdata
     .. attribute:: data
 
         Contains the resulting (sanitized) value of calling either of the
@@ -120,7 +120,7 @@ The Field base class
 
     .. attribute:: name
 
-        The HTML form name of this field. This is the name as defined in your 
+        The HTML form name of this field. This is the name as defined in your
         Form prefixed with the `prefix` passed to the Form constructor.
 
     .. attribute:: short_name
@@ -241,6 +241,8 @@ refer to a single input from the form.
                 image_data = request.FILES[form.image.name].read()
                 open(os.path.join(UPLOAD_PATH, form.image.data), 'w').write(image_data)
 
+.. autoclass:: MultipleFileField(default field arguments)
+
 .. autoclass:: FloatField(default field arguments)
 
    For the majority of uses, :class:`DecimalField` is preferable to FloatField,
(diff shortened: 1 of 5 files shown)


nblomqvist:
Would it be suitable to override the validation process of `MultipleFileField` such that validators do not need to be rewritten to also (and explicitly) handle multiple values?

davidism:
What validators need to be rewritten? You should be able to use `DataRequired` and `Length`, although I realize there's no test for it.

nblomqvist:
Sorry for being unclear. I was thinking of validators that people have written themselves, that are used for an input with a single value (e.g. a field holding a single `FileStorage` object), but that could just as well be applicable to multiple files. For example, I assume each validator in Flask-WTF will need to be rewritten to handle multiple files.

davidism:
Yes, that one's linked above and I'm assigned to it. 😉 I'm going to stick with rewriting that validator to work with either a single value or a list. I'm not sure what you have in mind for overriding `validate` that would allow the validator to work as-is. If you'd like to discuss it further, open a new issue.

nblomqvist:
Got it, thanks. I'll open an issue in the near future with an example on what I was thinking.

DuncanBetts:
I've been getting ImportError trying this:

```
from wtforms.fields import MultipleFileField
```

I've tried some other import statements, to no avail.

I've looked on Github Search, and can find no examples of anyone using this field type.

I've looked in the source code for tagged release 2.1, and can see MultipleFileField in the right place.

Any idea what I'm doing wrong? I'm hoping to use this along with Flask-Uploads, closest example of this I've found is http://www.patricksoftwareblog.com/tag/flask-uploads/ (with a FileField).

davidism:
This isn't released yet, not sure what code you're looking at.

DuncanBetts:
I read about the field here:

[WT Forms Documentation](http://wtforms.readthedocs.io/en/latest/fields.html#wtforms.fields.MultipleFileField)

davidism:
Those are the "latest" docs, i.e. the development docs. 2.1 is here: http://wtforms.readthedocs.io/en/2.1/fields.html#wtforms.fields.MultipleFileField. Unfortunately, I don't have maintainer access to the docs yet.

Jigsaw-Jams:
@davidism Do you have any idea when the latest version will be released? 

Also for the time being I'm hunting for a workaround. I've been doing: ```form.files(class_="form-control", multiple="") ``` but then I need to write filetype/extension validation in the view function.

 I tried writing a ```validate_files()``` method in the form class, but it only had access to the first file. Same with ```FileAllowed``` from Flask-WTF (it only validates against the first file being uploaded).

Overall main goal is to have a single ```<input type="file" multiple>``` that gets validated when running validate_on_submit(), I want to remove all the logic involved in checking the file extension from the view function.

If anyone has any ideas they would be greatly appreciated!

davidism:
As a temporary fix, you can always pull this patch into your own code, the fields don't have to be in the WTForms package to work.

As to the next release, I still have not received maintainer access for the majority of the tools (PyPI, Read the Docs, and org access on GitHub). I maintain a lot of Flask-related projects and have limited time, but will start focusing on this more once I have full access so that things go smoothly.

All I can do for now is merge PRs. Any help in the form of triage of existing issues and creating pull requests, tests, and docs is very helpful towards getting a new release ready.

r4jk3:
Whats the status of this bug?
Any breakthrough?

It would be nice to see MultipleFileField in action.

PR #282, opened 2016-06-22
Remove use of deprecated TextField and Required

davidism:
Change `TextField` to `StringField` and `Required` to `DataRequired`.


davidism:
I think the coverage decrease is just from changing the proportion of testable code.

PR #285, opened 2016-07-13
Improve french translations

sblondon:
The patch improves two sentences in french.

Do you need the messages.mo too?

Cheers,
Stéphane


davidism:
Yes, it looks like the repo contains the .mo file next to the .po, would you generate and commit that please?

sblondon:
I updated wtforms.mo and commited it.

PR #288, opened 2016-07-21
fix SelectField: do not coerce None into a string

georgschoelly:


tests/fields.py
```diff
@@ -367,7 +367,7 @@ def test_text_coercion(self):
             form.a(),
             '''<ul id="a">'''
             '''<li><input id="a-0" name="a" type="radio" value="True"> <label for="a-0">yes</label></li>'''
-            '''<li><input checked id="a-1" name="a" type="radio" value="False"> <label for="a-1">no</label></li></ul>'''
+            '''<li><input id="a-1" name="a" type="radio" value="False"> <label for="a-1">no</label></li></ul>'''
```
georgschoelly: Instead of changing the result, we could also pass `default=False` to the `RadioField` constructor.

collisdigital:
We hit this exact issue where we had a radio field whose name and value was `"None"` e.g. 

```Please select:
- None
- Low
- Medium
- High
```

WTForms couldn't distinguish between the user selecting the "None" option and not selecting an option at all and also meant that the None option was pre-selected in the radio field before the user had interacted with it. Pretty horribly bug!

We've worked around it with a custom coerce: https://github.com/ONSdigital/eq-survey-runner/issues/1015

But ideally this should be fixed in WTForms directly; what's the status on this one?

tests/fields.py
ftm: Why would it not be appropriate?

wtforms/fields/core.py
```diff
@@ -456,6 +456,10 @@ def iter_choices(self):
 
     def process_data(self, value):
         try:
+            # protect against coercing None,
+            # such as in text_type(None) -> "None"
+            if value is None:
+                raise ValueError()
             self.data = self.coerce(value)
```
ftm: I don't like how this is raising an error just to be caught, would it not be better to do
```python
try:
    self.data = None if value is None else self.coerce(value)
except (ValueError, TypeError):
    self.data = None
```

tests/fields.py
georgschoelly: There's no reason to select the second choice:

```
make_form(a=RadioField(choices=[(True, 'yes'), (False, 'no')], coerce=coerce_func))
```

Having none of the options pre-selected seems to be the better choice.

wtforms/fields/core.py
georgschoelly: I have no strong opinion on this.

ftm: In that case I'll change it when I merge it since I have to do it manually

alswl:
Any plan to update pypi version? @ftm

ftm:
@alswl I do not control when new versions are released. Please stop posting these comments on random issues and PRs, if you want to discuss this then please open a new issue.

Issue #172, opened 2015-03-10
Email validator accepts addresses with a trailing dot in local part


warvariuc:
```python
>>> import wtforms

>>> wtforms.validators.Email().regex.match('a.@gmail.com')
>>> <_sre.SRE_Match at 0x7f3fb8bf5030>

>>> wtforms.__version__
>>> '2.0.2'
```
> RFC 5322 allows having dots in the local part, however, there are two important limitations: It does not allow two or more consecutive dots, and it also does not allow the local part to start or end with a dot.

http://stackoverflow.com/questions/16010617/trailing-dot-in-local-part-of-email

warvariuc:
Just noticed the comment in the source code:
```
class Email(Regexp):
    """
    Validates an email address. Note that this uses a very primitive regular
    expression and should only be used in instances where you later verify by
    other means, such as email activation or lookups.
```
But I don't understand the value of this validator then.

kmmbvnr:
B/c fully RFC compatible validating email regexp a bit complex for everyday use =)

    (?:(?:\r\n)?[ \t])*(?:(?:(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(?:\r\n)?[ \t]
    )+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ \t]))*"(?:(?:
    \r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+(?:(?:(
    ?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|"(?:[^\"\r\\]|\\.|(?:(?:\r\n)?[ 
    \t]))*"(?:(?:\r\n)?[ \t])*))*@(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\0
    31]+(?:(?:(?:\r\n)?[ \t])+|\Z|(?=[\["()<>@,;:\\".\[\]]))|\[([^\[\]\r\\]|\\.)*\
    ](?:(?:\r\n)?[ \t])*)(?:\.(?:(?:\r\n)?[ \t])*(?:[^()<>@,;:\\".\[\] \000-\031]+
[...]

PR #294, opened 2016-08-27
improve Email validator

alanhamlett:
Uses the same `user_part` regex from [Django](https://github.com/django/django/blob/12b4280444b58c94197255655e284e4103fe00a9/django/core/validators.py#L170).

Related to #172, #169, and #117.

The new tests added here previously would fail, now they pass.

### DIFF
--- a/tests/validators.py
+++ b/tests/validators.py
@@ -34,6 +34,9 @@ class ValidatorsTest(TestCase):
         self.assertRaises(ValidationError, email(), self.form, DummyField('foo@bar'))
         self.assertRaises(ValidationError, email(), self.form, DummyField('foo@bar.ab12'))
         self.assertRaises(ValidationError, email(), self.form, DummyField('foo@.bar.ab'))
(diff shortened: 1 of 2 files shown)


alanhamlett:
Want to bring this up again and see if it can get merged?

illume:
This is a good test case to add as well: "b@example.com, b@example.com"
I've seen that one fall through quite a few validators. Pretty sure you patch catches it though :)

alanhamlett:
@crast could we get this merged?

kblomqvist:
Also related to #336?

alanhamlett:
@kblomqvist yes, this PR would fix #336 also.

asyncee:
Is this project alive? Can we get this PR merged?

alanhamlett:
Thanks!

davidism:
Note that this won't get a release until I get access to PyPI and Read The Docs from @crast.

alanhamlett:
It's just nice to not have to worry about rebasing now, in case another PR causes a merge conflict.

PR #298, opened 2016-09-17
Adds a validator check when creating a field

adrcad:
When doing `*Field(*args, validators=[DummyValidator])`, there is no check
on the callable received (`DummyValidator`). For builtin validators, if
this is not an instance, you get an incomprehensible error from
`_run_validation_chain`. This commit adds a check at field's
initialization time to ensure either a callable instance object or a
function has been given as a validator by raising a meaningful error
message.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import datetime
 import decimal
 import itertools
+import inspect
 
 from copy import copy
 
@@ -104,6 +105,8 @@ class Field(object):
         self.name = _prefix + _name
         self.short_name = _name
         self.type = type(self).__name__
+
+        self.checkValidators(validators)
         self.validators = validators or list(self.validators)
 
         self.id = id or self.name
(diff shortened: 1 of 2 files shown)


wtforms/fields/core.py
```diff
@@ -154,6 +157,14 @@ def __call__(self, **kwargs):
         """
         return self.meta.render_field(self, kwargs)
 
+    @classmethod
```
adrcad: classmethod decorator needed to be able to perform a check even when instantiating an unbounded Field (access through field_class)

ftm:
I like this idea, if you can fix the merge conflict and match the project code style (i.e. use `check_validators` not `checkValidators`) I'm up for merging it

wtforms/fields/core.py
```diff
@@ -154,6 +157,14 @@ def __call__(self, **kwargs):
...
+    def checkValidators(cls, validators):
```
davidism: Use snake_case names: `check_validators`.

wtforms/fields/core.py
```diff
@@ -154,6 +157,14 @@ def __call__(self, **kwargs):
...
+        if validators is not None:
+            for validator in list(validators):
```
davidism: Why wrap with `list` here?

wtforms/fields/core.py
```diff
@@ -154,6 +157,14 @@ def __call__(self, **kwargs):
...
+                if inspect.isclass(validator) or not hasattr(validator, '__call__'):
```
davidism: `if callable(validator)` and not inspect.isclass(validator):`

davidism: Perhaps also split these checks and have different messages: "{} is not a valid validator because it is not callable." and "{} is not a valid validator because it is a class, it should be an instance."

ftm:
I've completed the requested changes in a new PR #410 which I've merged

PR #299, opened 2016-10-01
Make sure a call to escape() doesn't crash script

willwac:
Refactored escape() calls to call _escape_wrapper() which uses a try clause to catch the TypeError raised when using html.escape() and calls escape() without the quote arg.

### DIFF
--- a/wtforms/widgets/core.py
+++ b/wtforms/widgets/core.py
@@ -13,6 +13,15 @@ __all__ = (
     'TextInput', 'Option'
 )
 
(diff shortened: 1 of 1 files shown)


davidism:
I don't see where this is crashing. `escape` is already being called with `quote` in all cases, and even if it wasn't it has a default value for `quote`. Python 2.7, 3.5, and 3.6.dev all still take two args, so I'm not sure why you think the second has been removed.  If you're still having this issue, please add a minimal example demonstrating the failure.

Issue #301, opened 2016-10-18
Do not escape HTMLStrings in SelectField choices


sonovice:
If choices are passed to `SelectField` as `HTMLString`, the content should not be escaped.

### Example
```python
from wtforms.widgets.core import HTMLString

languages = [('en', HTMLString('<i class="en flag"></i>English')),
             ('de', HTMLString('<i class="de flag"></i>Deutsch'))]
[...]
pref_language = SelectField(choices=languages)
```

### Possible solution
Instead of simply using `escape(...)` from `html` (like for example [here](https://github.com/wtforms/wtforms/blob/master/wtforms/widgets/core.py#L313)), use something like this:
```python
from html import escape

def escape_html(text_string):
    if hasattr(text_string, '__html__'):
        text_string = text_string.__html__()
    else:
        text_string = escape(text_string)
    return text_string
```

sonovice:
For some reason it still does not work in my project. Did I overlook a piece of code somewhere that would still escape my strings?

davidism:
You'll need to install the dev version of wtforms until a new release is made.  There is no schedule for a new release at this time.

sonovice:
I guess the problem is with flask-wtf. Since it's not compatible with wtforms >= 3.0, I have to work out something else... Thanks anyways, maybe this will come in handy sometime later.

PR #302, opened 2016-10-18
Choices in `SelectField` objects now handle `HTMLString` correctly.

sonovice:
Proposal for #301


davidism:
You have mistakenly committed your IDE and build files.  Please strip those commits and force push.

sonovice:
Thanks, I overlooked it.

davidism:
Please add another test to explicitly test `escape_html` in addition to testing it indirectly.

wtforms/widgets/core.py
```diff
@@ -14,6 +14,24 @@
...
+
+    If the optional flag quote is true (the default), the quotation mark
+    characters, both double quote (") and single quote (') characters are also
+    translated.
+
+    If a `HTMLString` is provied, it's assumed that whatever you give to
+    escape_html is a string with any unsafe values already escaped.
+    """
+    if hasattr(s, '__html__'):
+        s = text_type(s.__html__())
```
davidism: Remove `text_type`, as this will unmark a string that is already marked safe.

PR #310, opened 2016-12-07
Put field type and short_name in tr class for styleability

bryhoyt:


davidism:
If you need this, a custom widget should be fairly straightforward. The name part is especially problematic because names can easily be generic and interfere with other CSS rules.

Issue #311, opened 2016-12-13
Should not set fields as attributes on Form class


alanhamlett:
The fields are already kept in `self._fields` and setting them as attributes just causes name conflicts like [this one](https://github.com/flask-admin/flask-admin/issues/1128).

Why does [this line](https://github.com/wtforms/wtforms/blob/9be964158fbcd1af52b345451bbd14751127dd37/wtforms/form.py#L265) exist?

davidism:
This is pretty fundamental to how WTForms works. What would you want it to do instead? If they're not set, `form.my_field.data` wouldn't work.

alanhamlett:
Isn't that the same as `form.data['my_field']`? It's prone to errors to set fields on the class itself.

alanhamlett:
The only thing needed is to make `self._fields` available through a public property so you can access the whole field as you would `self.field_name`.

davidism:
I can't remove the attributes, code everywhere depends on them. `form.my_field.data` is an incredibly common pattern, I've actually never seen `form.data['my_field']` in general use.

alanhamlett:
I have a PR ready that caches the `form.data` dict. How about releasing this in a major version update since it's a breaking change?

From my experience, it's better to make breaking changes sooner rather than later in a project.

davidism:
I'm not willing to remove the attributes. Caching the data property is an unrelated optimization, not the blocker for this.

davidism:
Perhaps raising an exception if the attribute is already set and is not an `UnboundField`, to at least make the error obvious.

alanhamlett:
Ok, #312 does this by checking the field name against a restricted list before adding the attribute.

maxcountryman:
@davidism what do you think about providing a class, e.g. `SafeForm`, that subclasses `Form` and overloads `__init__` to work around issues like this?

PR #312, opened 2016-12-13
raise exception when using restricted field names, for #311

alanhamlett:
Related to #311.

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -155,6 +155,12 @@ class FormTest(TestCase):
         form = self.F()
         self.assertRaises(TypeError, form.__setitem__, 'foo', StringField())
 
+    def test_restricted_field_names(self):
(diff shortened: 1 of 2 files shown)


ftm:
Is there any way to do this without hardcoding the restricted values in? Perhaps seeing if the attribute already exists in the form?

ftm:
Due to the merge conflicts caused by recent changes to the repo and the age of this pull request I'm going to close it

alanhamlett:
> Due to the merge conflicts caused by recent changes to the repo and the age of this pull request I'm going to close it

I'll open a new PR with a better solution. What do you think about: Set form fields on a dict attribute, then override `Form.__getattr__`, and possibly `Form.__getattribute__` for better error messages, to check that dict attribute before calling super to raise an exception.

PR #313, opened 2016-12-15
Maxlength is added automatically to html tag of StringField

sblondon:
If the Length validator is used at the init, maxlength attribute will be included in the generated html tag.
If max value is not setted, the maxlength attribute is not added to the html tag.

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -136,6 +136,8 @@ class FiltersTest(TestCase):
 class FieldTest(TestCase):
     class F(Form):
         a = StringField(default='hello', render_kw={'readonly': True, 'foo': u'bar'})
+        b = StringField(default='world', validators=[validators.Length(max=42)])
+        c = StringField(default='world', validators=[validators.Length(min=12)])
 
     def setUp(self):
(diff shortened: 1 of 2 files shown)


sblondon:
This new patch add the same feature for the minlength attribute which is available in html5:
https://www.w3.org/TR/html5/forms.html#the-maxlength-and-minlength-attributes

ftm:
We already add the `required` attribute for the `InputRequired` validator so this seems like a good extension of that idea. If it can be made to pass the CI and fits in with the new project structure I'd be happy to merge it

ftm:
This should be done in [`validators.py`](https://github.com/wtforms/wtforms/blob/master/wtforms/validators.py), (see what we do for the `required` attribute and `InputRequired()`[here](https://github.com/wtforms/wtforms/blob/master/wtforms/validators.py#L220))

davidism:
We'll need to expand how flags work since right now they're just booleans.

ftm:
Oh yeah, I thought they were just strings that got added in but it doesn't look like it

PR #314, opened 2016-12-20
Add support for HTML in attribute values

### LINKED ISSUE #259: Attributes cannot contain escape sequences
I have a textarea field and would like to use a multi-line placeholder with it.  I'm using the `&#xA;` escape to get the newline, but wtforms escapes the `&`.  Example template (Jinja2):
```
{{ form.address(rows=3, cols=24, required=true,
	placeholder="123 Main road&#xA;SomeTown, XX 12345") }}
```
A solution to this would be to use [MarkupSafe](https://pypi.python.org/pypi/MarkupSafe)'s `escape` function if it is available.  This way I could simply use Jinja2's `|safe` filter on the text to turn it into `Markup` -- which the escape function would skip.

ShadowNinja:
This allows you to pass in raw HTML from libraries like MarkupSafe, while still escaping regular strings normally.
Fixes #259.

### DIFF
--- a/wtforms/widgets/core.py
+++ b/wtforms/widgets/core.py
@@ -68,7 +68,7 @@ def html_params(**kwargs):
         elif v is False:
             pass
         else:
-            params.append('%s="%s"' % (text_type(k), escape(text_type(v), quote=True)))
+            params.append('%s="%s"' % (text_type(k), escape_html(v)))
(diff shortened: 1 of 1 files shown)


ftm:
Can you give a situation in which you would want to do this? Which attributes require un-escaped HTML in them?

PR #325, opened 2017-03-01
New italian translation

corrador:
Hi, I created a new Italian translation because in the previous one lacked some messages.
I also corrected some errors and improved some messages.


corrador:
Also fix a problem similar to [#317](https://github.com/wtforms/wtforms/pull/317)
The Italian word 'invalido' is reserved for person with disabilities.
[https://it.wiktionary.org/wiki/invalido](https://it.wiktionary.org/wiki/invalido)

PR #330, opened 2017-04-13
add Unique validator

tysonholub:
This PR adds Unique validator to validate a fields uniqueness without enforcing a unique constraint at the database level. Consider precedence in Rails [validates_uniqueness_of](https://apidock.com/rails/ActiveRecord/Validations/ClassMethods/validates_uniqueness_of)


davidism:
Thanks for the PR, but this doesn't belong in WTForms, which is database agnostic. This is already present in WTForms-Alchemy. https://wtforms-alchemy.readthedocs.io/en/latest/validators.html#unique-validator

tysonholub:
Ah, that makes sense. I didn't know about that package, thanks!

PR #333, opened 2017-04-13
Update Email() validator in validators.py

ape364:
stripping email


davidism:
The validator definitely shouldn't modify the field data It would make more sense to modify the regular expression to skip surrounding whitespace. However, this should be done with a filter.

~~~python
StringField(filters=[lambda x: (x or '').strip()], validators=[Email()])
~~~

PR #342, opened 2017-06-19
adding has_error on form

lucasrc:
When in some case custom validation is necessary to know if the form is consistent can not access the form.errors as this has a great impact on how the rest of the script will work.


davidism:
I don't understand the use case for this. `if form.errors` should be sufficient.

davidism:
Closing. I think you were going for a "stop at the first error instead of processing the whole thing" optimization with the `any` call, but I don't think that's needed here because if there are errors you probably want the list anyway, which accessing `form.errors` would cache.

PR #343, opened 2017-06-25
Remove unnecessary step attribute

r3turnz:
The html 5 number inputs always render the step attribute with the default value.


davidism:
Need to fix the test.

davidism:
fixed test, added changelog in 6858b505db63536c7d7922dcda4bf7b35e853457

PR #345, opened 2017-06-27
update email validator regex. add whitespace email tests

tysonholub:
this PR fixes Email validator allowing email values with leading/trailing whitespace.


davidism:
I want to add the ability to strip whitespace in general, so I don't think it belongs here. For now, you can add a "validator" before the email validator that just sets `self.email.data` to the stripped version.

Issue #336, opened 2017-04-27
Wrong Email validation 


FadhelC:
There's an edge case where a wrong email address validates:
`foo@bar@baz.com` passes validation where it should't.
Only a single "**at sign**" should be present in an email address.

I would suggest the use of the following regular expression (derived from [RFC 5322](https://www.ietf.org/rfc/rfc5322.txt)) as the email validation regex:
 `r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)"`

davidism:
closed by #294

PR #353, opened 2017-09-06
Replace the regex used for email validation

Gowee:
The previous regex for email validation cannot handle some cases, such as the one in #336. So taking a new one is necessary.

As a library designed to work with web applications, it is reasonable to be compliant with [the email validity criterion in HTML5 standard](https://www.w3.org/TR/html5/forms.html#e-mail-state-(type=email)).

FYI, this change:
- does limit the maximum length of an email address to be 255 characters as is specified in RFC5322 and endorsed by many browsers;
- less strict (under some cases) than the previous one, for example, `user@a` is considered to be valid;
- support IDN by transforming a Unicode domain into IDNA before matching against the regex;
- expected to behave in the same way as HTML5 email validation.


alanhamlett:
I think this is already handled with #294?

PR #355, opened 2017-09-19
StringField shouldn't blank data during process_formdata.

### LINKED ISSUE #291: process_formdata on empty valuelist on StringField
On `StringField` class the method `process_formdata` assign `data` to empty string if `valuelist` is empty.
I think it is a wrong behavior as if an object was given and also a formdata, the value is replaced if the field is not in the formdata. Also all other field classes don't assign value on empty valuelist.

nanocell:
If StringField blanks data during process_formdata it incorrectly
discards the *data* supplied through FromField.process().

Fixes #291.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -529,8 +529,6 @@ class StringField(Field):
     def process_formdata(self, valuelist):
         if valuelist:
             self.data = valuelist[0]
-        else:
-            self.data = ''
 
     def _value(self):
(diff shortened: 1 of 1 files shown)


davidism:
Related to #280

davidism:
I need to rebase this to the 2.2-maintenance branch. I'm going to open a new PR with your commits to avoid messing up your fork. In the future, be sure to open PRs from a branch rather than your master so that maintainers can force push without messing up your fork.

PR #358, opened 2017-10-23
add zipcode validator

tysonholub:
PR adds a zipcode validator using regexes sourced from http://unicode.org/cldr/trac/browser/tags/release-26-0-1/common/supplemental/postalCodeData.xml?rev=11071

### DIFF
--- a/wtforms/validators.py
+++ b/wtforms/validators.py
@@ -530,6 +530,197 @@ class HostnameValidation(object):
         return True
 
 
+class Zipcode(object):
+    """
+    Validates a field as zipcode
+    """
+
+    # regexes sourced from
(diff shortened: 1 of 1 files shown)


davidism:
Thank you, but I don't like validating things with regexes, and I don't plan on adding more specific validators than what are already here.

tysonholub:
I would point out that your email validator uses regex (I have an open PR to fix issues with it). Zipcodes are far simpler to validate than email _and_ the expressions have already been defined by unicode.org. Why not make the library more robust?

davidism:
Yes, I don't like the email validator either, but that's already there. There is no defined stopping point for adding validators, and every piece of code added is something I need to support indefinitely. You may be interested in https://wtforms-components.readthedocs.io/en/latest/.

tysonholub:
I'll look into it. Thanks!

Issue #130, opened 2014-10-02
Required


nickretallack:
Why won't wtforms render the `required` attribute?  Instead everyone has to [hack it in using macros](https://github.com/mbr/flask-bootstrap/blob/master/flask_bootstrap/templates/bootstrap/wtf.html#L36).

Also, why can't I just say `wtforms.StringField(required=True)`?  `wtforms.StringField(validators=[wtforms.validators.DataRequired()])` is pretty verbose for something you have to do all the time.  It'd be nice to have a shortcut.

PR #361, opened 2017-11-12
Use HTML5 `required` attribute for required fields

yggi49:
If accepted, this PR will add the HTML5 boolean `required` attribute to all `<input>`, `<textarea>`, and `<select>` fields that have the `required` flag set (e.g., if the `InputRequired` validator is used).

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -136,6 +136,7 @@ class FiltersTest(TestCase):
 class FieldTest(TestCase):
     class F(Form):
         a = StringField(default='hello', render_kw={'readonly': True, 'foo': u'bar'})
+        b = StringField(validators=[validators.InputRequired()])
 
     def setUp(self):
         self.field = self.F().a
@@ -203,6 +204,10 @@ class FieldTest(TestCase):
         self.assertEqual(f1.items.choices, [('a', 'a')])
         self.assertEqual(f2.items.choices, [('b', 'b')])
         self.assertTrue(f1.items.choices is not f2.items.choices)
+
+    def test_required_flag(self):
+        form = self.F()
+        self.assertEqual(form.b(), u'<input id="b" name="b" required type="text" value="">')
 
 
 class PrePostTestField(StringField):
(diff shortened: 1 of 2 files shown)


AlexScheller:
Is there a discussion somewhere regarding this new behavior? It seemed from issue #130 and [this stack overflow answer](https://stackoverflow.com/a/33308165) that the prevailing opinion was to leave it up to the developer to choose whether to render the required attribute or not.

davidism:
It's now up to the developer to choose *not* to render it. 😉 `required=False` will force it not to render, or you can make a widget that doesn't set it. The `required` attribute prevents a request/response cycle for validation, which is valuable enough to make it the default. You can also prevent the behavior without messing with a lot of code by rendering the [`novalidate`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-novalidate) attribute on the `form` tag (or the `formnovalidate` attribute on a submit button).

My goal with the 2.2 release was to get it out, and I didn't really focus on documentation. If you'd like to make a PR that adds the reasoning above to the docs, that could be really helpful.

AlexScheller:
opened a PR basically just copying your response with a bit of editing. The docs built alright on my end. Hopefully this is similar to what you were looking for.

miguelgrinberg:
@davidism I would have preferred to make this feature opt-in, not opt-out. While the client-side validation works generally well, it was mentioned to me several times that users of apps that use Flask-Babel to offer several languages find it odd that these client-side validation popups cannot be translated and are in the language of the browser and not the currently selected language for the application.

href:
I agree that this should have been opt-in and not opt-out. Especially since there doesn't seem to be a straight-forward way to opt out. `novalidate` applies to the whole form and not just the required attribute. Presumably, there are other types of validation that you might still want.

davidism:
I'd be happy to review a PR that adds an attribute to the `Meta` class to control the policy for this. But I don't plan to change the default.

davidism:
I'm also open to brainstorming a more extensible "layout backend" + "render plugin" design for the entire syste. For example, you might have a "Boostrap" layout backend, or "HTML5 Required" and "ARIA" plugins.

davidism:
There also appears to be an API for modifying a browser's validation messages, although there would need to be some support from WTForms or an extension to make using it easier: https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/Form_validation#Customized_error_messages

davidism:
Turns out WTForms already has the ability to control how a form's fields are rendered. Override a form's `Meta` with a different `render_field` method.

```python
class MyForm(Form):
    class Meta:
            render_kw.setdefault('required', False)
            super().render_field(field, render_kw)
```

So I think at this point I'm inclined to leave things as-is.

PR #366, opened 2017-12-15
Add ValidationError StopValidation in __all__

lepture:
They are missing from `__all__`.


lepture:
wow, @davidism has so many accesses.

PR #369, opened 2018-01-11
Fix for BooleanField always evaluating to True when used with `Form(data=<dict>)`

nfvs:
When used in conjunction with Flask-WTF, `Field.process()` will be passed a formdata object (MultiDict), with a bool value `False`, even when used with `Form(data=obj)` or `Form(data=obj)`, which should support a raw dict.

Because this will then trigger `Field.process_formdata([False])`, `BooleanField.process_formdata()` will evaluate to `True`, which is, err, false. This PR correctly handles that case.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -696,7 +696,7 @@ class BooleanField(Field):
         ``('false', '')``
     """
     widget = widgets.CheckboxInput()
-    false_values = ('false', '')
+    false_values = (False, 'false', '')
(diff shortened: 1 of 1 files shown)


nfvs:
In flask-wtf, `Form.wrap_formdata` is the method that returns a json-based formdata, which is not handled by BooleanField:

https://github.com/lepture/flask-wtf/blob/master/flask_wtf/form.py#L64

PR #380, opened 2018-04-09
Fix #290 to broaden false_values for BooleanField

### LINKED ISSUE #290: Include 'False' as possible false value for boolean field
https://github.com/wtforms/wtforms/blob/b94260329964844d66c426072c649e64b6924a7c/wtforms/fields/core.py#L696

When submitting a form in a unit test using the flask client, if you do this:

```
client.post("/", data={"field": False})
```

The `False` value will become the string `"False"`, which does not register as `False` in a BooleanFIeld.

This would be fixed by adding `"False"` to line 696.

emin63:
As mentioned in #290, it would be nice if BooleanField was a little broader in what it accepts as a false value. Otherwise, when a client does a POST request using some other form of false value it gets ignored causing subtle bugs.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -693,10 +693,13 @@ class BooleanField(Field):
     :param false_values:
         If provided, a sequence of strings each of which is an exact match
         string of what is considered a "false" value. Defaults to the tuple
-        ``('false', '')``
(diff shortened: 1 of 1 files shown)


davidism:
See #290 for why I'm closing this.

PR #382, opened 2018-04-16
Adds how to populate a Fieldlist with enclosed Formfields to the docs

Razkaroth:
Clarifies how to populate FormFields inside FieldLists aided with a simple example of a carousel gallery. When I first strumbled across this particular mechanic I spent the next couple hours looking how this should be done. I ended up in this [stack question](https://stackoverflow.com/questions/29514798/filling-wtforms-formfield-fieldlist-with-data-results-in-html-in-fields).

But I really think this should be covered in the official documentation as it works differently than the normal populating mechanic.

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -413,6 +413,36 @@ complex data structures such as lists and nested objects can be represented.
             first_name  = StringField()
             last_name   = StringField()
             im_accounts = FieldList(FormField(IMForm))
+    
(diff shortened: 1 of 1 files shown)


davidism:
I'm not clear why you would be populating the form manually like this. Populate it the same way you populate any form: by passing in form data with the correct names to the form constructor.

PR #383, opened 2018-04-17
Added IPList Validator

TD4B:
Added an IPList validator that can handle comma delimited IP address values passed into the data.field. Figured I would just create a separate class rather than adding this into IPAddress.


TD4B:
seems to only pass with TagListField() 

`class TagListField(Field):
    """
    A text field, except all input is coerced to an integer.  Erroneous input
    is ignored and will not be accepted as a value.
    """
    widget = widgets.TextInput()

    def __init__(self, label=None, validators=None, **kwargs):
        super(TagListField, self).__init__(label, validators, **kwargs)

    def _value(self):
        if self.data:
            return u', '.join(self.data)
        else:
            return u''

    def process_formdata(self, valuelist):
        if valuelist:
            self.data = [x.strip() for x in valuelist[0].split(',')]
        else:
            self.data = []`

TD4B:
closed.

Issue #239, opened 2016-01-03
"aria_" should be handled like "data_" in html_params


MM1nd:
The title says it. Naturally this can be done by changing  

```
elif k.startswith('data_') 
```

to 

```
elif k.startswith('data_') or k.startswith('aria_')
```

but in the long run probably a more sophisticated system might be needed. Maybe a sophisticated implementation already exists.

Related Issue (https://github.com/wtforms/wtforms/issues/48)

Further reference if needed: http://www.w3.org/TR/wai-aria/

Alex

hammygoonan:
+1

Further reference here as well: [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA)

This is a feature I'd really like to see implemented.

It's now part of Foundation's standard help-text formatting as well [http://foundation.zurb.com/sites/docs/forms.html#help-text](http://foundation.zurb.com/sites/docs/forms.html#help-text)

crast:
Good idea. I'm putting it on the schedule for 3.0

ftm-1:
@crast Is this feature anywhere near to being included? I have a PR ready to submit if this is still something you guys want?

davidism:
@ftm-1 I'm the new maintainer. If you have a PR, please submit it! :-)

ftm:
Hi @davidism, I was @ftm-1 but now I'm on this account (I managed to get a better username). I have reopened the PR on my new account so it should be ready to go.

davidism:
Closed in #389

PR #387, opened 2018-05-05
Modified html_params to treat aria attributes the same as data attributes

ftm-1:
Fixing #239, when specifying field attributes in a template if the attribute begins with `aria_`, then the generated attribute shall begin with `aria-`, just like how `data` attributes are handled.

For example
```python
{{ form.username(aria_describedby='usernameHelp') }}
```
will now make
```html
<input aria-describedby="usernameHelp" id="username" name="username" value="" type="text">
```

I also corrected the documentation for `html_params`, clarifying how when using `data_` or `aria_` attributes every underscore gets replaced with a hyphen - not just the first one.


ftm-1:
Closing this as I've switched to a different GitHub account so I'm going to re-open the PR from that one

PR #394, opened 2018-06-04
update documentation for "required" flag in DataRequired and InputRequired Validators

AlexScheller:
This PR updates the documentation regarding the "required" flag in the DataRequired and InputRequired Validators. As of Version 2.2, this flag causes the "required" attribute to render in forms. This update to the documentation reflects this and notes the behavior it will result in, as well as providing a few methods of overriding this behavior. The body of this update comes from an answer provided by @davidism in PR #361 with some editing by me.


davidism:
Thanks!

PR #403, opened 2018-06-11
fix for ipv6 validator

### LINKED ISSUE #385: IPAddress validator return False for valid IPv6 address.
Hello,

the address "2001:718:1C01:1111::" is valid IPv6 address, however, the IPAddress validator returns False for it.

jirivrany:
Here is a fix for ipv6 validator. It requires ipaddress package to be installed in Python 2. Python 3 is ok. I kept the old tests and added several more in pytest style. 

closes #385

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -49,15 +49,14 @@ class ValidatorsTest(TestCase):
         self.assertRaises(ValidationError, equal_to('foo'), self.form, DummyField('different_value'))
 
     def test_ip_address(self):
-        self.assertEqual(ip_address()(self.form, DummyField('127.0.0.1')), None)
-        self.assertRaises(ValidationError, ip_address(), self.form, DummyField('abc.0.0.1'))
-        self.assertRaises(ValidationError, ip_address(), self.form, DummyField('1278.0.0.1'))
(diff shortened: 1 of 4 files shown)


jirivrany:
I'm not sure why the build for Python 2.7 is failing? Travis log shows, that ipaddress package has been installed sucessfully. Perhaps tox should be used for this external dependency configuration?

tests/test_validators.py
```diff
@@ -250,3 +250,46 @@ def test_any_of(self):
...
+
+@pytest.mark.parametrize("address", [
+    u"147.230.1000.25",
+    u"2001:718::::",
+])
+def test_ip4address_raises(address):
+    adr = ip_address(ipv6=True)
+    field = DummyField(address)
+    with pytest.raises(ValidationError):
+        adr(None, field)
```
ftm: Why are all of these outside the `test_ip_address` function?

jirivrany: That's because of the `test_ip_address` is method written in UnitTest style. And @davidism told me,    that I can use the PyTest style for the new tests. Because the original tests should be rewritten to PyTest style too in the future. And so I use the PyTest parametric tests.

davidism: This is fine.

ftm:
So is this all ready to merge then?

PR #412, opened 2018-06-13
Implement support of PEP 440 versioning by means bumpversion.

vald-phoenix:
Here's in this PR I implemented versioning according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#pre-releases).
We'll get separation on four sections: alpha, beta, release candidate and final release by means such pattern: `{major}.{minor}.{patch}.{release}{n}`.
```
X.Y.ZaN   # Alpha release
X.Y.ZbN   # Beta release
X.Y.ZrcN  # Release Candidate
X.Y.Z     # Final release
```
The workflow for the version 3.0.0.a0 will be looking somehting like this:
```sh
bumpversion patch # 3.0.0.a0 → 3.0.1.a0
bumpversion n # 3.0.1.a0 → 3.0.1.a1
bumpversion release # 3.0.1.a1 → 3.0.1.b0
bumpversion release # 3.0.1.b0 → 3.0.1.rc0
bumpversion release # 3.0.1.rc0 → 3.0.1
bumpversion minor # 3.0.1 → 3.1.0.a0
bumpversion major # 3.1.0.a0 → 4.0.0.a0
bumpversion n # 4.0.0.a0 → 4.0.0.a1
bumpversion release --tag # 4.0.0.a1 → 4.0.0.b0 it will create for us a tag v4.0.0.b0
```
P.S. I'd like to consult with you where is it sould be placed inside documentation and when these commands should be fired as well because you better understand workflow of WTForms.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,20 @@
+[bumpversion]
+current_version = 3.0.0.a0
+commit = True
+tag = False
+files = setup.py docs/conf.py wtforms/__init__.py
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\.(?P<release>[a-z]+)(?P<n>\d+))?
+serialize = 
+	{major}.{minor}.{patch}.{release}{n}
+	{major}.{minor}.{patch}
+
+[bumpversion:part:release]
+optional_value = gamma
+values = 
+	a
+	b
+	rc
+	gamma
+
+[bumpversion:part:n]
+
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,9 +54,9 @@ copyright = '2008 by the WTForms team'
 # other places throughout the built documents.
 #
 # The short X.Y version.
-version = '3.0'
+version = '3.0.0.a0'
 # The full version, including alpha/beta/rc tags.
-release = '3.0dev'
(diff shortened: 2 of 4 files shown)


vald-phoenix:
Yes, you're right about a release tag. So I guess it'll be better to come up only with a dev option that increments, right? E.g.:
```sh
bumpversion patch # 3.0.0 → 3.0.1.dev0
bumpversion n # 3.0.1.dev0 → 3.0.1.dev1
bumpversion release # 3.0.1.dev1 → 3.0.1
bumpversion patch # 3.0.1 → 3.0.2.dev0
```
About workflow, I meant, if somebody will create a PR, should that person patch version or only core maintainers will update a version of the project?

davidism:
After releasing a version, I bump to the next version too, so everything that happens on master represents the next feature release, everything that happens on a maintenance branch represents the next bugfix release. Contributors wouldn't bump versions.

To be honest the current process isn't that complex. The only thing bumpversion would automate is setting the release version, tagging, then setting the next dev version. It might actually require more typing to use bumpversion than to do this manually.

vald-phoenix:
So we don't need this one?

davidism:
I'm going to close this for now. Thanks for looking into it, it's definitely interesting. I'll keep it in mind and think about how we can improve the release process.

PR #413, opened 2018-06-14
Rewrite tests pytest part 1.

jirivrany:
test_validators and test_webob has been rewritten to PyTest layout


davidism:
Should `common.py` become `conftest.py`?

davidism:
Did you split the classes out into functions and parametrize them? Is every test case still represented? Possibly a good way to verify this is to check that coverage is the same or better.

jirivrany:
Yes, I split the classes and use parametrized test everywhere it was possible. The overall coverage increased. 
New version:
```
tests/test_validators.py ...........................................................................................................................  [ 91%]
tests/test_webob.py ...                                                                                                                               [ 92%]
```
Current master branch
```
tests/test_validators.py .........................                                                                                                    [ 86%]
tests/test_webob.py ...                                                                                                                               [ 88%]
```
Make `conftest.py` from `common.py` is a good idea. I just must test it, in case that it will broke test files which are not converted yet.

PR #417, opened 2018-06-22
Simple pep8 line length checks and typo fix

whb07:
Just a quick and simple break in lines longer than 88 chars which is the standard for Black. Also renamed a variable 'l' to 'length' to be more explicit. Lastly, a simple typo fix in a docstring.

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -897,11 +897,13 @@ class FormField(Field):
         self._obj = None
         if self.filters:
             raise TypeError(
-                "FormField cannot take filters, as the encapsulated data is not mutable."
+                "FormField cannot take filters, "
+                "as the encapsulated data is not mutable."
             )
         if validators:
             raise TypeError(
-                "FormField does not accept any validators. Instead, define them on the enclosed form."
+                "FormField does not accept any validators. "
+                "Instead, define them on the enclosed form."
             )
 
     def process(self, formdata, data=unset_value):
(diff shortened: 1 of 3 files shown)


wtforms/validators.py
```diff
@@ -376,7 +379,8 @@ def __call__(self, form, field):
 
 class IPAddress(object):
     """
-    Validates an IP address. Requires ipaddress package to be instaled for Python 2 support.
+    Validates an IP address. Requires ipaddress package
```
davidism: Line length for docstrings is 72.

wtforms/fields/core.py
```diff
@@ -991,7 +995,8 @@ def __init__(
         super(FieldList, self).__init__(label, validators, default=default, **kwargs)
         if self.filters:
             raise TypeError(
-                "FieldList does not accept any filters. Instead, define them on the enclosed field."
+                "FieldList does not accept any filters."
+                " Instead, define them on the enclosed field."
```
davidism: Be consistent with space before / after continuation. I prefer this after continuation style because it makes it more clear that there's a continuation, and feels more like the binary operator rule.

davidism: Prefer wrapping strings at 72 to match docstrings for now.

whb07: Great catches !  I didn’t know that about the docstrings, but going forward I know Black’s default line length is 88. So do you want to set it at that or...?

davidism: PEP 8 specifies different lengths for docstrings vs code. Use 72 for docstrings, 88 for code.

davidism: Let's use the space after continuation for the reasons I gave above.

davidism:
I wanted to force push to this PR branch, but since you based this off your master, doing this would mess up your fork. Always use a branch when creating a PR.

whb07:
Yeah my bad. 😡 I accidentally noticed this too late after I made a secondary branch with some changes. I’ll close this for now and make a new pr

PR #422, opened 2018-06-23
Test widgets

whb07:
This PR changes the unittest format to a Pytest format for test_widgets.py. In order to do so, a couple changes were made:

in `conftest.py`
- added basic_widget_dummy_field, select_dummy_field, html5_dummy_field. They are fixtures for a specific instance of the DummyField class that was used within each class

- Added more attributes to DummyField in order to accommodate being used by the test_widgets tests.

in `test_widgets.py`

- removed the wildcard import in favor for the more explicit imports.
- changed most assertions being the `assertEqual` for Pytest's `assert`

in `validators.py`

- changed the `object.attribute` access in favor for the `getattr()` to allow for a default value when the label.text value is being accessed. 

All tests passing.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,6 +19,21 @@ def dummy_field_class():
 
 
 @pytest.fixture
+def basic_widget_dummy_field(dummy_field_class):
+    return dummy_field_class("foo", name="bar", label="label", id="id")
+
+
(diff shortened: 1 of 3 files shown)


davidism:
Squashed. Removed change to `EqualTo`, fixed the test instead.

PR #427, opened 2018-06-27
Fix a bug in email validation

vald-phoenix:
I found tricky, small bug. If a value will be something like this: `foo@bar.ua foo@bar.ua`, it'll be valid, what's absolutely wrong.
`user_part` & `domain_part` will be split in the next way:
```python
user_part = "foo@bar.ua foo"
domain_part = "bar.ua"
```

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -374,7 +374,7 @@ class Email(object):
         if message is None:
             message = field.gettext("Invalid email address.")
 
-        if not value or "@" not in value:
+        if not value or "@" not in value or " " in value:
             raise ValidationError(message)
 
         user_part, domain_part = value.rsplit("@", 1)
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -239,6 +239,7 @@ def test_valid_email_passes(email_address, dummy_form, dummy_field):
         "foo.@bar.co",
         "foo@foo@bar.co",
         "fo o@bar.co",
+        "foo@bar.ua foo@bar.ua",
     ],
 )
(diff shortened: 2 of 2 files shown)


davidism:
Fairly sure that local part can contain space and ampersat. Regex needs fixing.

alanhamlett:
That regex comes from Django, and they haven't updated it in 3 years:
https://github.com/django/django/blob/11b8c30b9e02ef6ecb996ad3280979dfeab700fa/django/core/validators.py#L164

They also have a lot of [wontfix tickets](https://code.djangoproject.com/search?q=EmailValidator&noquickjump=1&changeset=on&milestone=on&ticket=on&wiki=on), so looks like it's up to us. Here's some related material:

[RFC 5322 compliant regex](https://stackoverflow.com/a/201378/1290627)
[Wikipedia Local-part Semantics](https://en.wikipedia.org/wiki/Email_address#Common_local-part_semantics)

Would it be better to just use a robust library built for email validation, like [python-email-validator](https://github.com/JoshData/python-email-validator), instead of regex?

davidism:
Yes, I would prefer using email_validator like we did with ipaddress.

vald-phoenix:
@davidism so we're going to use the third-party library, right?

davidism:
Yeah, let's do that.

vald-phoenix:
@davidism okay, I'll do it on weekend.

alanhamlett:
@vald-phoenix had a spare 16 mins today and wanted to feel productive... hope you don't mind that I implemented it: #429

davidism:
closing for #429

PR #428, opened 2018-07-01
remove \r in ending of lines.

itamarjp:


davidism:
I'm not sure I understand why this was necessary, but I'm going to be regenerating the Spinx files anyway, so I'll deal with it then.

PR #429, opened 2018-07-05
use email_validator library for Email validator instead of regex

alanhamlett:
Implements outcome of conversation in #427 using [email_validator](https://pypi.org/project/email_validator/) to validate email address.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,10 @@ setup(
     include_package_data=True,
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     setup_requires=["Babel>=2.6.0"],
-    install_requires=["MarkupSafe", "email-validator"],
-    extras_require={"locale": ["Babel"]},
+    install_requires=["MarkupSafe"],
+    extras_require={
+        "ipaddress": ["ipaddress"],
+        "email": ["email_validator"],
+    },
     cmdclass=command_classes,
 )
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ setup(
     setup_requires=["Babel>=2.6.0"],
     install_requires=["MarkupSafe"],
     extras_require={
-        "ipaddress": ['ipaddress;python_version<"3"'],
+        "ipaddress": ['ipaddress;python_version<"3.3"'],
         "email": ["email_validator"],
     },
     cmdclass=command_classes,
--- a/setup.py
+++ b/setup.py
@@ -99,6 +99,9 @@ setup(
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     setup_requires=["Babel>=2.6.0"],
     install_requires=["MarkupSafe"],
-    extras_require={"ipaddress": ["ipaddress"], "email": ["email_validator"]},
+    extras_require={
+        "ipaddress": ['ipaddress;python_version<="2.7"'],
+        "email": ["email_validator"],
+    },
     cmdclass=command_classes,
 )
--- a/setup.py
+++ b/setup.py
@@ -98,7 +98,7 @@ setup(
     include_package_data=True,
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     setup_requires=["Babel>=2.6.0"],
-    install_requires=["MarkupSafe"],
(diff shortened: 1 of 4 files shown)


src/wtforms/validators.py
```diff
@@ -339,50 +340,31 @@ def __call__(self, form, field, message=None):
...
-                \\[\001-\011\013\014\016-\177]
-            )*"  # quoted-string
-        )$
-        """,
-        re.IGNORECASE | re.VERBOSE,
-    )
+    defaults = {"check_deliverability": False}
 
-    def __init__(self, message=None):
+    def __init__(self, message=None, **options):
```
davidism: `options` needs a mention in the docstring.

src/wtforms/validators.py
```diff
@@ -2,6 +2,7 @@
 
 import re
 import uuid
+from email_validator import validate_email, EmailNotValidError
```
davidism: Should this be optional like `ipaddress`?

alanhamlett: Maybe, but I prefer just adding a hard dependency since email validation is very common. What are the benefits of making it optional?

davidism: If you don't use email validation (I don't for many projects), you now have an extra dependency that you don't use. I'd prefer an optional dependency since email validation isn't required to making WTForms work.

I don't recall exactly why, but I think using CC0 for a license is potentially problematic, so forcing users to install email_validator could cause licensing issues as well.

alanhamlett: Made optional with a2a7afe5d3dc7712ab949a512ba13ead40eed779. Also, added extras for both `email_validator` and `ipaddress` dependencies so they can be installed at the same time as `wtforms`. Removed `locale` extra since `Babel` is already a required dependency, so there's no reason for having it in extras.

setup.py
```diff
@@ -98,7 +98,10 @@ class BDistWheel(CompileCatalogMixin, BaseBDistWheel):
     include_package_data=True,
     python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     setup_requires=["Babel>=2.6.0"],
-    install_requires=["MarkupSafe", "email-validator"],
-    extras_require={"locale": ["Babel"]},
+    install_requires=["MarkupSafe"],
+    extras_require={
+        "ipaddress": ["ipaddress"],
```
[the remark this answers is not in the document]
alanhamlett: Extras are just ways to define optional dependencies, so it's up to the user to install wtforms with various extras or not depending on what they need and their environment. Not sure if we can specify extras only available for Python 2.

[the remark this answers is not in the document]
alanhamlett: I'll see if this works:

https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies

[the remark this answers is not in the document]
davidism: I think this would work:

```
'ipaddress:python_version < "3"': ["ipaddress"],
```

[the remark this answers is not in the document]
alanhamlett: It works! Tested 50a0c3c96a88f3e35abf8c6146731c5b675c596a on Python 2.7 and Python 3.6 both with `pip install -e .[ipaddress]` and `ipaddress` was only installed on Python 2.7.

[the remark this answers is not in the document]
alanhamlett: Oh, didn't see your last comment. I'll update to `< 3`.

setup.py
```diff
@@ -100,7 +100,7 @@ class BDistWheel(CompileCatalogMixin, BaseBDistWheel):
     setup_requires=["Babel>=2.6.0"],
     install_requires=["MarkupSafe"],
     extras_require={
-        "ipaddress": ['ipaddress;python_version<="2.7"'],
+        "ipaddress": ['ipaddress;python_version<"3"'],
```
davidism: Uhh, not to drag this out too long, but technically this should be "<3.3". :sweat: Then again, we don't support <3.4, so... :man_shrugging:

alanhamlett: Updated with 325367f6469dce804f73b7891b0f64254ac37ec1. 👍

alanhamlett: Not sure if you've seen the latest push, and don't want to merge without full review.

alanhamlett:
Should be ready to merge now.

One thing to consider is improving [install docs](https://wtforms.readthedocs.io/en/stable/crash_course.html#download-installation) to mention extras, but that could be another PR.

PR #441, opened 2018-08-15
Fixed if vaule is None validate failed

zengqiu:
In ``process_data``, if value is None, don't coerce to a value, but validate will failed.

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -535,7 +535,7 @@ class SelectField(SelectFieldBase):
 
     def pre_validate(self, form):
         for v, _ in self.choices:
-            if self.data == v:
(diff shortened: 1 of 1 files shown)


src/wtforms/fields/core.py
```diff
@@ -535,7 +535,7 @@ def process_formdata(self, valuelist):
 
     def pre_validate(self, form):
         for v, _ in self.choices:
-            if self.data == v:
+            if self.data == v or self.data is None:
```
vald-phoenix: You can, even more, simplify this stuff like this  `if self.data == v or not self.data`.

ftm:
This is a reasonably old PR which causes test failures and currently causes merge conflicts so I will close it for now. If you still want it to be considered please re-open it.

PR #451, opened 2018-11-08
IntegerField should check type when process data

whusnoopy:
Core fields like `IntegerField` will check type on `process_formdata` by reload from base `Field`, but not check on `process_data` in current version. If we pass unsupported type in `data` for `Form`, like a string 'x', the form won't raise any error and will pass the origin string as value for user, it doesn't make sense for this case.

As `BooleanField` did in `process_data`, `IntegerField` should do the same thing, so I add the progress and test case for `IntegerField`.

Maybe other fields in `fields/core.py` like `FloatField` should do type check in `progress_data` also.

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -666,6 +666,16 @@ class IntegerField(Field):
         else:
             return ""
 
+    def process_data(self, value):
+        if value is not None and value is not unset_value:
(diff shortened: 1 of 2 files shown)


ftm:
Thank you for adding this fix, apologies for not accepting the PR for over a year.

PR #452, opened 2018-11-14
IntegerField Validation

yashpokar:
IntegerField validation doesn't provides user friendly message, Code for handling that was there already but don't know why It wasn't triggered. I haven't debugged it.


ftm:
`.isnumeric()` is not a good way to do this as it does not work with negative numbers.

```python
>>> "123".isnumeric()
True
>>> "-123".isnumeric()
False
```

PR #454, opened 2018-12-13
Raise sensical error on Field.choices being None

seanpianka:
Many people have run into issues where a Field's `choices` field is unset or set to None, yet a non-descriptive TypeError is currently raised, which leads to StackOverflow posts like this: https://stackoverflow.com/questions/15936764/im-having-problems-with-wtforms-selectfields-when-i-use-a-post-with-flask

The following change will give the developer a hint of what the issue could be.

Edit: It looks like the CI build that failed was a code-style check?

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -534,6 +534,9 @@ class SelectField(SelectFieldBase):
                 raise ValueError(self.gettext("Invalid Choice: could not coerce"))
 
     def pre_validate(self, form):
+        if self.choices is None:
+            return TypeError(self.gettext("Choices cannot be None"))
+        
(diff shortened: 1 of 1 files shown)


ftm:
Apologies for the delay in getting to this, this seems like a common enough issue that it warrants its own error message so thanks for contributing it.

As you point out, the CI is failing on code style. WTForms uses [Black](https://github.com/python/black) for code formatting, please make sure that your code fits this format by running Black on it.

ftm:
It's been over six months with no changes/fixes so I'm going to close this PR for now. I have raised an issue regarding the unhelpful error message and will pick it up in a future change.

PR #456, opened 2018-12-23
Split Form.validate

omarryhan:
Split Form.validate into Form.validate and _get_extras for better sub-classing.

### DIFF
--- a/src/wtforms/form.py
+++ b/src/wtforms/form.py
@@ -279,15 +279,17 @@ class Form(BaseForm, metaclass=FormMeta):
             else:
                 super().__delattr__(name)
 
-    def validate(self):
-        """
-        Validates the form by calling `validate` on each field, passing any
-        extra `Form.validate_<fieldname>` validators to the field validator.
-        """
+    def _get_extras(self):
         extra = {}
         for name in self._fields:
             inline = getattr(self.__class__, "validate_%s" % name, None)
             if inline is not None:
                 extra[name] = [inline]
+        return extra
 
-        return super().validate(extra)
+    def validate(self):
+        """
+        Validates the form by calling `validate` on each field, passing any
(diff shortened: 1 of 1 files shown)


ftm:
Apologies for not getting to this for so long.

Your PR is failing the style check part of the CI. WTForms uses the [Black code style](https://github.com/psf/black). If this is still something you're interested in having included into WTForms please fix the code style.

azmeuk:
@omarryhan I rebased your patch to get the latest changes.

This patch would require tests involving that `_get_extras` method, else you won't have a guarantee that a future refactoring won't break this.
What do you think?

omarryhan:
Sure.

I pulled the branch that you rebased but I couldn't run the tests, because of this error:

```
  File "/home/omar/.local/lib/python3.6/site-packages/setuptools/wheel.py", line 156, in _convert_metadata
    os.rename(dist_info, egg_info)
PermissionError: [Errno 13] Permission denied: '/home/omar/wtforms/.eggs/pytz-2019.3-py3.6.egg/pytz-2019.3.dist-info' -> '/home/omar/wtforms/.eggs/pytz-2019.3-py3.6.egg/EGG-INFO'
```

It looks like an issue with my Windows Subsystem Linux. Unfortunately, this is the only environment I am able to work on right now.

Initially, I wanted to create the extra method in the PR to make this lib work with https://github.com/omarryhan/sanic-wtf, but then I figured that it wasn't necessary. So feel free to close this pull request if you don't feel like writing tests for this PR.

Thanks

davidism:
Thanks for working on this and following up. I'm going to close it, I don't think we need to split it apart. The standard way to override `Form.validate` isn't affected by this:

```python
class MyForm(Form):
    def validate(self):
        result = super().validate()

        if not result:
            return False

        return my_other_validation
```

Also, #584 allows passing `extra_validators` to `validate`.

Issue #462, opened 2019-02-02
wtforms.validators.HostnameValidation does not allow underscores in the domain portion of the URL


camillser:
Hey, there seems to be a problem with the way the wtforms.validators.HostnameValidation is handling the underscore character. Underscores should be allowed in domain names per [1]. This problem was found using wtforms.validators.URL and that validator depends on wtforms.validators.HostnameValidation.

Reproducible example:
```
import wtforms.validators
validator = wtforms.validators.HostnameValidation()
validator("a_b.example.org")
```

The result of the above code is that validator returns False. The expected result is True.

[1] https://tools.ietf.org/html/rfc2181#section-11

ftm:
Closing issue as it has been fixed by the PR referenced above

PR #463, opened 2019-02-02
Allow underscores in wtforms.validators.HostnameValidation.

camillser:
Example patch for #462

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -609,7 +609,7 @@ class HostnameValidation(object):
     This is not a validator in and of itself, and as such is not exported.
     """
 
-    hostname_part = re.compile(r"^(xn-|[a-z0-9]+)(-[a-z0-9]+)*$", re.IGNORECASE)
+    hostname_part = re.compile(r"^(xn-|[a-z0-9_]+)(-[a-z0-9_]+)*$", re.IGNORECASE)
     tld_part = re.compile(r"^([a-z]{2,20}|xn--([a-z0-9]+-)*[a-z0-9]+)$", re.IGNORECASE)
 
     def __init__(self, require_tld=True, allow_ip=False):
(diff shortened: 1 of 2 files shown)


src/wtforms/validators.py
```diff
@@ -609,7 +609,7 @@ class HostnameValidation(object):
     This is not a validator in and of itself, and as such is not exported.
     """
 
-    hostname_part = re.compile(r"^(xn-|[a-z0-9]+)(-[a-z0-9]+)*$", re.IGNORECASE)
+    hostname_part = re.compile(r"^(xn-|[a-z0-9_]+)(-[a-z0-9_]+)*$", re.IGNORECASE)
```
davidism: Is a leading `_` valid, even though a leading `-` isn't?

camillser: From some casual research it looks like leading and trailing underscores are valid for domain names [1], but underscores are not valid in hostnames [2]. I think the problem here is that the URL validator is using HostnameValidation to verify the domain name when it should be using a more permissive validator. I see that HostnameValidation is not exported so I think the intention of that validator was to be a domain name validator and to be used exclusively in URL validator. I hope that makes sense. Let me know if it would be a good idea to change the name of HostnameValidation to make the intentions of the class clearer.

[1] https://tools.ietf.org/html/rfc2181#section-11
[2] https://www.ietf.org/rfc/rfc1123.txt

davidism: Yeah, I'm fine limiting it to domain names.

ftm:
Was it decided what the outcome of this was going to be? Should we rename `HostnameValidation` to something like `DomainNameValidation` and make it more permissive like the PR is doing?

davidism:
I'm fine with making the validation more permissive. It's not a public API, and it's only used by `URL` validator, so maybe we should just deprecate the class and inline the code. Related to #474, although whatwg-url doesn't validate the domain.

ftm:
I'll merge it for now then, even if later we end up inlining it.

davidism:
@ftm thanks! Remember to add a changelog entry

PR #465, opened 2019-02-09
options for Regexp to validate raw data

kaiwensun:
If the field is of non-string type, eg. `FloatField` and `IntegerField`, we may still want to use Regex to validate the input. One potential use case is to validate a price in `FloatField` type to match `^\d*(\.\d{1,2})?$`

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -320,6 +320,9 @@ class Regexp(object):
         `regex` is not a string.
     :param message:
         Error message to raise in case of a validation error.
+    :param use_raw:
+        Validate the raw data from `formdata` instead of `data`.
+        (Default False)
     """
 
     def __init__(self, regex, flags=0, message=None, use_raw=False):
@@ -352,7 +355,7 @@ class Email(object):
 
     :param message:
         Error message to raise in case of a validation error.
-    :param granular_messsage:
+    :param granular_message:
         Use validation failed message from email_validator library
         (Default False).
     :param check_deliverability:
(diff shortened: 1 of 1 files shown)


src/wtforms/validators.py
```diff
@@ -322,14 +322,16 @@ class Regexp(object):
...
+    def __init__(self, regex, flags=0, use_raw=False, message=None):
         if isinstance(regex, string_types):
             regex = re.compile(regex, flags)
         self.regex = regex
+        self.use_raw = use_raw
         self.message = message
 
     def __call__(self, form, field, message=None):
-        match = self.regex.match(field.data or "")
+        data = field.raw_data[0] if self.use_raw else field.data
```
kaiwensun: question to admins: Instead of validating only the first raw_data, maybe we need to test the regex against all items?

kaiwensun:
@davidism Would you mind taking a look at this PR?

ftm:
Would it not be simpler to just ensure that any data is converted into something that the regex can handle before attempting to match on it.

For example just changing:
https://github.com/wtforms/wtforms/blob/534cca0fb8947415edf44229b45f4dc4bf8c4fb7/src/wtforms/validators.py#L332
to
```python
match = self.regex.match(str(field.data or ""))
```

davidism:
If it's a price, it should be a `DecimalField`, not a `FloatField`, at which point there's already params for number of places. Validating the shape of arbitrary data is probably better done with specific validators rather than regex.

I don't think the example given is compelling enough to justify converting *back* to a string just to run a validator that probably has a better alternative. *Maybe* the validator could catch a `TypeError` then try `raw_data`, but again I'm not sure this is necessary.

PR #474, opened 2019-03-05
extended  URL() validator for cases of basic-auth.

BOPOHA:
extended  URL() validator for cases of basic-auth.

according to RFC1738, the syntax for the scheme-specific data:

`//<user>:<password>@<host>:<port>/<url-path>`

https://tools.ietf.org/html/rfc1738#section-3

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -495,7 +495,14 @@ class URL(Regexp):
     """
 
     def __init__(self, require_tld=True, message=None):
-        regex = r"^[a-z]+://(?P<host>[^/:]+)(?P<port>:[0-9]+)?(?P<path>\/.*)?$"
+        regex = (
+            r"^[a-z]+://"
+            r"(?P<login>[a-zA-Z0-9_-]+(?P<password>:[a-zA-Z0-9_-]*)?@)?"
+            r"(?P<host>[^/:]+)"
+            r"(?P<port>:[0-9]+)?"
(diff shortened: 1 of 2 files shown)


yuralar:
This patch will solve the [issue](https://pagure.io/copr/copr/issue/525) in Fedora Copr Build Service.

ftm:
Apologies for the delay in looking at this, many thanks for raising this issue and providing a potential fix.

I'm not sure why this wasn't originally included, perhaps it was because where HTTP URLs are defined in [Section 3.3 of the same RFC](https://tools.ietf.org/html/rfc1738#section-3.3) they don't include a username/password section:
```
   An HTTP URL takes the form:

      http://<host>:<port>/<path>?<searchpart>
```

However, since there is nothing else to suggest this validator is solely for use with HTTP URLs, I would like to merge this.

My only concern is that your regex does not seem to cover all the possibilities outlined in the RFC for the characters in the username/password field. [Section 5](https://tools.ietf.org/html/rfc1738#section-5) has a formal BNF-like definition of a URL including which characters are allowed in which part. Would it be possible for you to modify your regexes and tests to fit this definition? I figure if we're going to follow the RFCs we should probably follow them completely? @davidism what do you think?

davidism:
Like the `Email` validator, I don't like that this uses regex. Like `Email`, we should consider an optional external library: https://pypi.org/project/whatwg-url/

PR #481, opened 2019-05-05
Add a BooleanRequired validator for use with BooleanField

alswl:
provide BooleanRequired for BooleanField

idea from: https://code.luasoftware.com/tutorials/python/wtforms-booleanfield-value-and-validation/

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -649,6 +649,26 @@ class HostnameValidation(object):
         return True
 
 
+class BooleanRequired(object):
+    """
+    Validates that boolean input was provided for this field.
+    """
+    field_flags: Tuple[str] = ('required',)
+
+    def __init__(self, message: Optional[str] = None):
+        self.message = message
+
+    def __call__(self, form: Form, field: BooleanField):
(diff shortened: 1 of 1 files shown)


alswl:
WIP

ftm:
Can you explain the purpose of this because at the moment I'm struggling to see one.

The article you link says that BooleanField does not accept `False` as a false value, when it [does](https://github.com/wtforms/wtforms/blob/a4df965494232d3292abbfcf0498148be358f0a8/src/wtforms/fields/core.py#L790) and has done [since January 2018](https://github.com/wtforms/wtforms/commit/cbcb0c81ad561dd9ab1149417074ae7d0dd49318) (one month before the article was released) although admittedly it looks like the docs were not updated when that change was made.

From what I can see your validator then attempts to fix this non-issue by ensuring the raw data can **only** be `True` or `False` which defeats the whole point of accepting a range of false values (i.e. not just `True` or `False`).

Additionally you have not written any tests for the validator.

ftm:
As there has been no response, I am closing this.

Issue #460, opened 2019-01-10
Booleanfield broken in 2.2


hyperknot:
Booleanfield's checked state got broken in 2.2

If an object's member variable is True, the generated field's data will be False. 

```python
# user.verified is True

class UserEditForm(Form):
    verified = BooleanField('verified')

form = UserEditForm(obj=user)
print(user.verified) # True
print(form.verified.data) # False
```

ftm:
I can't seem to verify your example using WTForms version 2.2.1. Are you still having this error?

```python
>>> from wtforms import Form, BooleanField
>>> class User:
...     verified = True
... 
>>> user = User()
>>> user.verified
True
>>> class UserForm(Form):
...     verified = BooleanField('verified')
... 
>>> form = UserForm(obj=user)
>>> form.verified.data
True
```

ftm:
It's been nearly three months since I asked if the error is still occurring and there has been no response so I'm going to assume this is no longer an issue.

hyperknot:
@ftm yes it's still present! 

I've downgraded that production site to < 2.2 since I've opened this ticket, for a workaround. Just now I've got your message and I tried it locally, and the bug is still present in 2.2.1!

hyperknot:
I made a minimal reproducible case. The key is request.POST parsing, that triggers the bug in 2.2 and 2.2.1.

```
from wsgiref.simple_server import make_server

from pyramid.config import Configurator
from wtforms import BooleanField, Form


class User:
    verified = None


class UserForm(Form):
    verified = BooleanField('verified')


def user_edit(request):
    user = User()
    user.verified = True

    form = UserForm(request.POST, obj=user)

    print(user.verified)
    print(form.verified.data)

    return {}


if __name__ == '__main__':
    with Configurator() as config:
        config.add_route('user_edit', '/')
        config.add_view(user_edit, route_name='user_edit', renderer='json')
        app = config.make_wsgi_app()
    server = make_server('0.0.0.0', 7000, app)
    server.serve_forever()
```

You only need to install `pyramid` for this and just open or curl `http://localhost:7000`.

ftm:
Ah, okay, thanks for providing that example. I can now verify this issue, I'll take a look.

ftm:
I believe I have found the issue but it's not actually to do with BooleanField, it's to do with how the lack of POST data is being interpreted incorrectly.

[...]

Issue #402, opened 2018-06-07
No way to tell that a given form is submitted


davidism:
Related to at least #119, #242, #267, #280, #291, #355, #401

We do `if formdata is not None` before calling `process_formdata` on each field in a form. This broke the rather common case where users were passing `request.form` to a form even if the method wasn't POST because `request.form` isn't `None`. Previously we checked `bool(formdata)`, which handled this case but caused processing to behave inconsistently depending on what was submitted.

Flask-WTF somewhat handles this by only passing `request.form` if the method is actually a submit method (POST, PUT, DELETE).

None of the approaches are really correct. Maybe you have a client that always sends some extra data (for example, an auth key). Or you have two forms on the same page, whichever was not submitted will still see the same non-empty form data as the one that was submitted.

```python
f1 = FormA(request.form)  # was submitted, sees data
f2 = FormB(request.form)  # was not submitted, sees same data
```

*Most* built-in fields mask this issue because they only change `data` (which is set from the default / object data first) if `formdata.getlist('name')` is not empty. However, `StringField` will set `data = ''` if nothing was submitted (which #355 changes). And `SelectMultipleField` and `MultipleFileField` will overwrite with an empty list.

This can also cause weird issues during validation with `InputRequired`. If there are two forms, and you do something like `if f1.validate(): ... elif f2.validate(): ...`, and `f1` was not submitted, it will show a "input required" error even if a default value was being displayed, which is confusing.

I'm not sure what the solution is here, but am open to discussion. We can break things, I don't care about backwards compatibility for 3.0.

CC @lepture @alanhamlett

davidism:
This is more broadly important if we want to validate other data sources such as JSON, or PATCH requests, which do not have the same constraints and behaviors as HTML forms.

davidism:
We could check if any of the form keys match any of the field names before processing. But this is expensive and unnecessary for the most common case of a single form. It's also incorrect for validation, where "no form data" might legitimately be a validation issue.

Alternatively, we could have a `Form.is_submitted(formdata)` method which just returns `True` by default, or `True` if any keys match submit field names and values. Flask-WTF already has this concept, although it's not implemented the way I just described, and it only applies to validation right now, not processing.

davidism:
[...]

PR #486, opened 2019-05-05
WIP: Fix incorrect processing of WebOb-style form data

ftm:
As detailed in my comment on issue #460, (see https://github.com/wtforms/wtforms/issues/460#issuecomment-489435199), when wrapping `formdata` it appears frameworks that use WebOb (such as Pyramid) in the case of a submission containing no form data (such as a GET request) populate their `request.POST` field with a placeholder `NoVars` object instead of just `None` like some other frameworks. WTForms interpreted this as valid (but empty) form submission data.

This caused issues further down the line for BooleanFields. When an HTML form is submitted, if it contains unchecked checkboxes they aren't included in the request body sent to the server. So BooleanField can't tell if it really was an empty form submission, or a form containing an unticked box, and assumes it was an unticked box and set its `data` to `False` erroneously.

This pull request introduces an extra check when wrapping `formdata` to verify the length of the form data. If it is zero-length (like `NoVars`) then it just returns `None` so the rest of the processing behaves correctly.

### DIFF
--- a/src/wtforms/meta.py
+++ b/src/wtforms/meta.py
@@ -37,13 +37,21 @@ class DefaultMeta(object):
         :param formdata: Form data.
         :return: A form-input wrapper compatible with WTForms.
         """
+
         if formdata is not None and not hasattr(formdata, "getlist"):
             if hasattr(formdata, "getall"):
+                # Presence of getall attribute implies Webob-style multidict
+                if len(formdata) == 0:
+                    # If length is 0, formdata is NoVars: it doesn't contain any
+                    # information so None should be returned to prevent errors
+                    # occuring due to formdata not being None
+                    return None
+
                 return WebobInputWrapper(formdata)
             else:
                 raise TypeError(
-                    "formdata should be a multidict-type wrapper that"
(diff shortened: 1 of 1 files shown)


src/wtforms/meta.py
```diff
@@ -37,13 +37,21 @@ def wrap_formdata(self, form, formdata):
         :param formdata: Form data.
         :return: A form-input wrapper compatible with WTForms.
         """
+
         if formdata is not None and not hasattr(formdata, "getlist"):
             if hasattr(formdata, "getall"):
+                # Presence of getall attribute implies Webob-style multidict
+                if len(formdata) == 0:
```
ftm: As @hyperknot pointed out in https://github.com/wtforms/wtforms/issues/460#issuecomment-489437421, checking for zero-length wouldn't necessarily be safe as this could break for valid but empty form submissions.

azmeuk:
@ftm This seems (again) related to #402 :) Any idea what to do with this PR?

azmeuk:
Closing as inactive.

PR #491, opened 2019-06-03
Wrapping number data in float

IsaacWarren:
I was trying to use number range validator and it crashed because of an invalid comparison between str, data, and a number, min/max. When I wrapped the data field in float it was fixed.


davidism:
If `data` is not already the correct type, there is something else wrong. If you believe this is an issue with WTForms and not your code, please open an issue with a minimal reproducible example and a full traceback.

PR #510, opened 2019-08-15
Fill FieldList data with matching objects data

azmeuk:
Let us edit data at the beginnig of a `FieldList` that has a min_entries, and provide a data object to fill empty indices:

```python
>>> import wtforms
>>> from tests.common import DummyPostData
>>> class F(wtforms.Form):
...     a = wtforms.FieldList(wtforms.TextField(), min_entries=5)
>>> a  = F().a
>>> pdata = DummyPostData({"a-1": "foo"})
>>> data = ["bar0", "bar1", "bar2"]
>>> a.process(pdata, data)
>>> a.data
```
Without the PR `a.data` is `['foo', '', '', '', '']`, which is unexpected because indices `1` and `2` are empty, where I need the to be `"bar1"` and `"bar2"`.

With this PR `a.data` is `['foo', 'bar1', 'bar2', '', '']`.

This pull request fixed #509 If you want, I can submit another pull request for the `master` branch.

Éloi

PR #511, opened 2019-08-15
Fill FieldList data with matching objects data - 2

azmeuk:
So this one is the following of #510, but as it changes a bit the behavior of `FieldList` I prefered to make it a separate pull request.

Let us edit data in the middle of a `FieldList` that has a `min_entries`, and provide a data object to fill empty indices:
```python
>>> import wtforms
>>> from tests.common import DummyPostData
>>> class F(wtforms.Form):
...     a = wtforms.FieldList(wtforms.TextField(), min_entries=5)
>>> a  = F().a
>>> pdata = DummyPostData({"a-1": "foo"})
>>> data = ["bar0", "bar1", "bar2"]
>>> a.process(pdata, data)
>>> a.data
```

So with #510 `a.data` would have took this value : `['foo', 'bar1', 'bar2', '', '']`, which is weird because `foo` was supposed to be at index `1` but now is at index `0`.

With this PR `a.data` is `['bar0', 'foo', 'bar2', '', '']`. `foo` is at index `1`, indices `0` and `2` come from `data` and indices `4` and `5` are empty because they are not initialized.

### DIFF
--- a/tests/fields.py
+++ b/tests/fields.py
@@ -809,6 +809,14 @@ class FieldListTest(TestCase):
         self.assertEqual(a.data, ['foo', 'flaf', 'bar'])
         self.assertRaises(AssertionError, a.append_entry)
 
+    def test_min_entries_default_values(self):
+        F = make_form(a=FieldList(self.t, min_entries=5, max_entries=5))
+        a = F().a
+        pdata = DummyPostData({"a-0": "foo"})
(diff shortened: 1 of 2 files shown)


azmeuk:
If you prefer this changes to happen in wtforms 3 instead of wtforms 2, I am totally OK to make my pull request against `master` branch.

PR #513, opened 2019-08-20
Rewrited tests to match pytest conventions

### LINKED ISSUE #408: Rewrite UnitTests to PyTest
@davidism mentioned [here](https://github.com/wtforms/wtforms/issues/385) that there is a plan to rewrite tests to be using PyTest layout instead of UnitTest. I have some spare time so I can start work on that if it helps to the project.

azmeuk:
This should probably fix #408 

At this point we could use test functions instead of classes of tests, but I prefer to discuss this before. Which one do you prefer?

### DIFF
--- a/tests/test_csrf.py
+++ b/tests/test_csrf.py
@@ -6,7 +6,6 @@ from functools import partial
 import hashlib
 import hmac
 import pytest
-from unittest import TestCase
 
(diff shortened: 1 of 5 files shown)


azmeuk:
@ftm I saw you were active lately on this repository. Do you have thoughts about this pull request?

Issue #154, opened 2015-01-18
WTForms 3.0


crast:
https://github.com/wtforms/wtforms/milestones/3.0

WTForms 3.0
===========

WTForms for a while has required a redesign to reflect the way things are going in how people actually use forms. WTforms 2.0 was a bridge version, a way to make a major version that hints at the upcoming changes but have it be compatible with the majority of people's existing code and form classes,while also letting people be aware of pending deprecation.

Overview
---------

* Proper support for input data in other forms than just form input. (ahem, JSON)
* Make it clearer the distinction between fields, data type coercion, and widgets in a way that doesn't give users headaches, and enables easy use.
  * Distinguish fields that only use one value, use zero or one value, or use multiple values, and allow compatibility to be had based on some of these facets.
  * Simplify the general use case for the vast majority of field implementations
  * Decouple the way fields access form values from the way that coercion happens, to allow easy implementation for third-party libraries
* Better and easier to use form composition tools, and multi-value fields.

Major Features
--------------

* Input instead of coming as a disparate collection of parameters, is now put into a specialty data class `FormInput`
  * FormInput holds and gates access to multiple input data types, including "models", defaults, JSON/dictionary input, and form data.
  * An API of clearly defined methods allows fields to get input data without caring as much about how this data comes and implementation details behind it
  * Having this clear API contract allows this object to be replaced with anything else satisfying the contract (better support for many other frameworks) and also behaviors
  such as defining the priority of what to do with missing data on a global level.
  * For maximum compatibilty, the old `formdata, obj, **kwargs` type data can still be supplied, but a FormInput will be created with these.
* Kill off SelectField, or make it less sucky
  * In wtforms 1&2, the biggest problem encountered was that having multiple Decimal values, for example, a SelectField with `coerce=Decimal` was not symmetric to all the capabilities provided by DecimalField
  * (Potential) If we could allow multiple choices for any field type, then it more strongly reinforces the idea of a field-per-datatype e.g. `DecimalField(choices=[...])` and so on.
* Make field enclosures less of a headache to use
  * FormField and FieldList are very capable, but there has to be a better way, especially in the fringe cases.

delijati:
WTForm + JSON + REST = https://github.com/kvesteri/wtforms-json

pbu88:
Hi, maybe this improvement can be included in 3.0 https://github.com/wtforms/wtforms/issues/185 Is not a real issue, just a feature request :)

lepture:
[...]

PR #520, opened 2019-10-28
Remove duplication of fields in core.py and htm5.py

rabinadk1:
Since everything is imported in `__init__.py` both form `core.py` and `html5.py`, some of the classes are imported twice.
Since `core.py` is imported first, the classes from `core.py` is the one selected when the duplicated classes are called. This results in `type= "text"` for `IntegerField` and `DecimalField` too.

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -660,7 +660,7 @@ class IntegerField(Field):
     is ignored and will not be accepted as a value.
     """
 
-    widget = widgets.TextInput()
+    widget = widgets.NumberInput()
 
     def __init__(self, label=None, validators=None, **kwargs):
         super(IntegerField, self).__init__(label, validators, **kwargs)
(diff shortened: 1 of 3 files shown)


azmeuk:
Indeed that feels wrong that the wildcard imports import different classes with the same name.
https://github.com/wtforms/wtforms/blob/129040ab4d5af83a05ddc7b63f854d48be3d55aa/src/wtforms/fields/__init__.py#L2-L4

I understand from #154 and #234 that the plan was to provide a generic way for "widget families", but I don't think this will happen soon. Meanwhile I am not sure what we should do those html5 fields. Use them by default, and keep the others for compatibility? Or keeping using them optionally?

A solution to fix those wildcard imports could be to rename either the html5 fields, or the *old* ones, so there would not be name conflicts.

azmeuk:
Duplicated code issues will be solved with #594

Issue #523, opened 2019-11-05
wtforms.validators.URL does not allow query parameters in the URL


laurihy:
Using `wtforms.validators.URL` to validate URL with query parameters incorrectly returns `False`.

```python3
import wtforms
 
class URLForm(wtforms.Form):
     url = wtforms.StringField('MyLink', validators=[wtforms.validators.URL()])
 
URLForm(url='https://example.com').validate()
# -> True

URLForm(url='https://example.com?this=fails').validate()
# -> False, but this should be True as it's a valid URL
```

Seems like this is because the regex used in URL validation _incorrectly_ includes query parameters to the hostname, and then `HostnameValidation` fails.

https://github.com/wtforms/wtforms/blob/master/src/wtforms/validators.py#L498

```python
import re
from wtforms.validators import HostnameValidation
url_regex = re.compile(r"^[a-z]+://(?P<host>[^/:]+)(?P<port>:[0-9]+)?(?P<path>\/.*)?$")
match  = url_regex.match('https://example.com?query=params')
print(match.group('host'))
# -> example.com?query=params
HostnameValidation(match.group('host'))
# -> False
```

Instead, the regex should separate hostname and query parameters, and only use the actual hostname for `HostnameValidation`.

I'll cook up a PR to fix this later this week, unless someone is faster than that :)

laurihy:
https://github.com/wtforms/wtforms/pull/524 should fix this

PR #524, opened 2019-11-05
URLs with query params should pass URL validator

laurihy:
Fixes https://github.com/wtforms/wtforms/issues/523

Previously URL validator didn't accept URLs with query parameters if there wasn't also a path specified. This was, because in that case the old regex would include query-params to the `host`, and attempt to call `HostnameValidation` with the incorrect `host`.

Old regex:
```python
regex = re.compile(r"^[a-z]+://(?P<host>[^/:]+)(?P<port>:[0-9]+)?(?P<path>\/.*)?$")
matches = regex.match('https://foo.io?no=path')
print(matches.group('host'))
# -> foo.io?no=path
```

New regex:
```python
regex = (
    r"^[a-z]+://"
    r"(?P<host>[^\/\?:]+)"
    r"(?P<port>:[0-9]+)?"
    r"(?P<path>\/.*?)?"
    r"(?P<query>\?.*)?$"
)
matches = regex.match('https://foo.io?no=path')
print(matches.group('host'))
# -> foo.io
print(matches.group('query'))
# -> ?no=path
```

Added a bunch of tests to verify this behavior.

### DIFF
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -382,6 +382,9 @@ def test_equal_to_raises(
         u"http://foobar.dk/",
         u"http://foo-bar.dk/",
         u"http://foo_bar.dk/",
+        u"http://foobar.dk?query=param",
+        u"http://foobar.dk/path?query=param",
+        u"http://foobar.dk/path?query=param&foo=faa",
         u"http://foobar.museum/foobar",
         u"http://192.168.0.1/foobar",
         u"http://192.168.0.1:9000/fake",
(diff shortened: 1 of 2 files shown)


tests/test_validators.py
```diff
@@ -400,7 +403,14 @@ def test_valid_url_passes(url_val, dummy_form, dummy_field):
...
+    "url_val",
+    [
+        u"http://localhost/foobar",
+        u"http://foobar",
+        u"http://foobar?query=param&foo=faa",
+    ],
+)
 def test_valid_url_notld_passes(url_val, dummy_form, dummy_field):
     """
     Require TLD option se to false, correct URL should pass without raising
```
whb07: can you change this to "TLD option set to false" so fix typo.

whb07:
looks good in general just missing the tweaks for the typo and an extra test case for the tld test set

laurihy:
Thanks for the good comments! Added a few more test cases, and fixed the typo in the comment.

whb07:
Great, I saw the changes and it looks good to me. I also played with it locally and manually to verify some behavior and the fixes work 🙃

PR #526, opened 2019-11-21
Shortcut syntax for choices with same label & data

dominickj-tdi:
This would allow for a convenient shortcut syntax for entering choices. If label and data are both the same, you can do `['One', 'Two', 'Three']` instead of `[('One', 'One'), ('Two', 'Two'), ('Three', 'Three')]`

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -526,7 +526,7 @@ class SelectField(SelectFieldBase):
             choices = self.choices
         else:
             choices = zip(self.choices, self.choices)
-            
+
         for value, label in choices:
             yield (value, label, self.coerce(value) == self.data)
 
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -522,7 +522,12 @@ class SelectField(SelectFieldBase):
         self.validate_choice = validate_choice
 
     def iter_choices(self):
-        for value, label in self.choices:
+        if isinstance(self.choices[0], (list, tuple)):
+            choices = self.choices
+        else:
+            choices = zip(self.choices, self.choices)
+            
+        for value, label in choices:
             yield (value, label, self.coerce(value) == self.data)
 
     def process_data(self, value):


ftm:
Your PR is failing the style check part of the CI. WTForms uses the [Black code style](https://github.com/psf/black). If this is still something you're interested in having included into WTForms please fix the code style.

ftm:
Thanks for sorting that, I'll merge it now

spookey:
I am afraid there is a bug in it..
In my application I am filling the choices of some select field with elements coming from a database query.

If the result is empty, the choices are effectively set to ``[]``.

So the following lines produce some ``IndexError`` in this case - there is no index 0 in an empty list:
```
def iter_choices(self):
        if isinstance(self.choices[0], (list, tuple)):
```

spookey:
```py
    def iter_choices(self):
        if isinstance(self.choices[0], (list, tuple)):
            choices = self.choices
        else:
            choices = zip(self.choices, self.choices)

        for value, label in choices:
            yield (value, label, self.coerce(value) == self.data)
```

It is possible to restore the original functionality by introducing some ``if self.choices:`` check before:

```py
    def iter_choices(self):
        if self.choices:
            if isinstance(self.choices[0], (list, tuple)):
                choices = self.choices
            else:
                choices = zip(self.choices, self.choices)

            for value, label in choices:
                yield (value, label, self.coerce(value) == self.data)
```

I am not sure if it is too dirty for a fix, please also note #572 :simple_smile:

dinosawrusrex:
I think the `pre-validate` function of `SelectField`needs to be updated similarly to `iter_choices`. Currently, `pre-validate` will break if `choices` was something like `['One', 'Two', 'Three']`. The culprit is the line `for v, _ in self.choices:`.

```py
def pre_validate(self, form):
    if self.choices is None:
        raise TypeError(self.gettext("Choices cannot be None"))

    if self.validate_choice:
        for v, _ in self.choices:
            if self.data == v:
                break
        else:
            raise ValidationError(self.gettext("Not a valid choice"))
```

A solution similar to @spookey's as [above](https://github.com/wtforms/wtforms/pull/526#issuecomment-617675939) would work. Happy to put together a solution and open a PR if no one is on this yet!

Issue #255, opened 2016-02-10
For IntegerField or FloatField when required the 0 should be an accepted value.


lipis:
and not allowing the empty value `''` but the 0 should be a valid number..

lipis:
@crast Why exactly is this invalid?!

```python
class MyForm(wtf.Form):
  amount = wtforms.FloatField('Amount', [wtforms.validators.required()])
  count = wtforms.IntegerField('Count', [wtforms.validators.required()])
```

In the above form, I can add any value to the `amount` and `count` except `0` or `None`.. which is a bit strange because `0` is perfectly accepted input when `required` and `None` should be not allowed.. since it's not a number.

If I have these field `optional` then of course 0 is allowed.. but it also accepts the empty string (or None) which is not a number and I'll have to do an extra action in order to validate it (even though I'm asking for Integer or Float)

lipis:
@crast ?!

NixonInnes:
I get this issue too. Why is "0" not a valid value for FloatField and IntegerField when the validator DataRequired is applied?

lipis:
@NixonInnes That is exactly my point :D `None` or `` or any non numerical string should be forbidden.. not `0` :)

georgekis:
+1 keep up

NixonInnes:
I've found a somewhat workaround, if you set the default value to 0.0 in the model and don't use any validators on the form definition; it will accept 0, and still error on "Not a valid float value".

crast:
@lipis this is a well-documented issue with the "required" validator,  there's two paragraphs of notes on it:  http://wtforms.readthedocs.org/en/2.1/validators.html#wtforms.validators.DataRequired

For back-compatibility reasons, we cannot change the behaviour of the "required"/"data_required" validator in the 2.x series, because it would be a breaking change and that is a violation of semantic versioning policy.

We do have another validator called [InputRequired](http://wtforms.readthedocs.org/en/2.1/validators.html#wtforms.validators.InputRequired) which works for integer 0 and float 0.0 because it checks the form input is specified, not the value truthiness.

[WTForms 3.0](https://github.com/wtforms/wtforms/issues/154) is a pretty hefty re-design and it addresses many common issues, and will address the "required" issue.

I'm reopening this and setting the milestone as 3.0 so that it's tracked

lipis:
Thanks @crast for clearing this up.

RazerM:
[...]

PR #528, opened 2019-11-28
InputRequired checks None instead of truthiness

adrian410:
If int = 0 then it's not recognized as a value and not pass the required validator

https://github.com/wtforms/wtforms/issues/255#issuecomment-559493117

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -298,7 +298,7 @@ class InputRequired(object):
         self.message = message
 
     def __call__(self, form, field):
-        if not field.raw_data or not field.raw_data[0]:
+        if not field.raw_data or field.raw_data[0] is None:
             if self.message is None:
                 message = field.gettext("This field is required.")
             else:


ftm:
Could you provide a more detailed title for the pull request?

Additionally, your changes cause some existing test cases to fail.

azmeuk:
Also, can you add some unit tests please?

davidism:
I don't understand the implications of this change, given the existing documented behavior of the two required validators. Recommend closing this if we don't receive the requested fixes or a response in a couple days.

adrian410:
The implication is to allow 0 as an acceptable value for IntegerField and FloatField. 

It is related to this issue : https://github.com/wtforms/wtforms/issues/255

I will try to find a better solution in the next days without the tests failing.

davidism:
That's just a restatement of the original comment on this PR. I'm talking about the implications of changing the behavior of this validator from "is not False" to "is not None". Remember, this is used for more than integers. The difference between `DataRequired` and `InputRequired` are already documented, and discussed at length in the linked issue.

PR #530, opened 2019-12-12
Added html5 input field type "month"

rambowtech:
There was already a widget defined,  but no field available. I added it, so the html5 input type "month" may be utilized.

First time I contribute on Github. I hope everything is correct. Feedback would be appreciated.


azmeuk:
Hello. Thank you for your contribution.
I suppose a new field would require some tests (to be sure it works, and that future contributions wont break it), and some documentation (maybe a reference in `docs/fields.rst` would be enough).

You can have a look on `DateField` tests in `tests/test_fields.py` for instance.

davidism:
Continued in #593

PR #532, opened 2019-12-17
Fix Travis CI config

ftm:
An improvement and fix to the CI config:

- Removed deprecated `sudo` key
- Pin `coverage` version in `py` Tox environments (see [comment below](https://github.com/wtforms/wtforms/pull/532#issuecomment-566705131))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ python:
   - pypy3
 env: TOXENV=py,coveralls
 
-matrix:
+jobs:
   include:
     - python: "3.8"
       env: TOXENV=stylecheck,docs-html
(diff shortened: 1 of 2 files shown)


ftm:
The CI is currently failing due to an interesting bug being caused by an incompatibility between the version of `coverage` used in the `py` and `coveralls` Tox environments.

In the `py` environments there is no specific version specified resulting in the latest version being pulled (version 5.0), whereas `coveralls` depends on a version <5.0. Version 5.0 introduced a new format to store the reports so when `coverage combine` gets ran in the `coveralls` environment with version 4.5.4 it crashes as it can't read the new v5.0 format reports.

PR #535, opened 2019-12-21
Fix simple typo: requred -> required

### LINKED ISSUE #534: Fix simple typo: requred -> required
There is a small typo in tests/test_validators.py.
Should read `required` rather than `requred`.

timgates42:
Closes #534


ftm:
Thanks for catching that

PR #536, opened 2019-12-25
fix SelectMultipleField to fix issue where initial data is always empty

Sets88:
If you have initial data in SelectMultipleField field SelectMultipleField.data will be overwritten by SelectMultipleField.process_formdata method because there is no check if valuelist is empty

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -574,14 +574,15 @@ class SelectMultipleField(SelectField):
             self.data = None
 
     def process_formdata(self, valuelist):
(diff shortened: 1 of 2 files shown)


ftm:
Your PR is currently failing CI. Please ensure that it is passing all tests. You may need to rebase your branch off of `master` to receive all of the latest updates.

davidism:
I don't think this is correct. If you don't send form data for a field, that field is supposed to be empty. Setting initial data is only for rendering, final data is only based on processed data.

PR #537, opened 2020-01-05
Add more helpful error message when SelectField.choices is None

### LINKED ISSUE #533: Unhelpful error message raised when SelectField.choices is None
As pointed out in the PR #454, the error raised if validation of a SelectField with no choices is attempted isn't especially helpful:

```python
from wtforms.fields.core import SelectField
from wtforms.form import Form

class F(Form):
    sf = SelectField("Test field")

form = F()

form.validate()
```

```
Traceback (most recent call last):
  File "SelectField_error.py", line 9, in <module>
    form.validate()
  File "[...]/python3.7/site-packages/wtforms/form.py", line 300, in validate
    return super(Form, self).validate(extra)
  File "[...]/python3.7/site-packages/wtforms/form.py", line 136, in validate
    if not field.validate(self, extra):
  File "[...]/python3.7/site-packages/wtforms/fields/core.py", line 244, in validate
    self.pre_validate(form)
  File "[..]./python3.7/site-packages/wtforms/fields/core.py", line 549, in pre_validate
    for v, _ in self.choices:
TypeError: 'NoneType' object is not iterable
```

A more preferable alternative would be a helpful error message that actually tells the user that the `choices` list is `None` or is otherwise unusable.

The solution #454 proposed of catching it at the `pre_validate` point is viable however I closed the PR as the author was not fixing their code or resolving conflicts.

bebleo:
fixes #533

I've added a more explicit message for when `SelectField.choices` is None and a couple of associated tests (one for when it is and one for when it's an empty list).

~~Additionally, the build on Travis-CI appears to be failing due to the release on December 30, 2019 of coveralls 1.10 which uses coverage 5.0.1. To work around this I've set the dependency in `tox.ini` to be coveralls <= 1.9.2~~ PR #541 renders this unnecessary.

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -545,6 +545,9 @@ class SelectField(SelectFieldBase):
                 raise ValueError(self.gettext("Invalid Choice: could not coerce"))
 
     def pre_validate(self, form):
+        if self.choices is None:
+            raise TypeError(self.gettext("Choices cannot be None"))
+
         if self.validate_choice:
             for v, _ in self.choices:
                 if self.data == v:
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -402,6 +402,20 @@ class SelectFieldTest(TestCase):
         self.assertEqual(len(form.a.errors), 1)
         self.assertEqual(form.a.errors[0], "Not a valid choice")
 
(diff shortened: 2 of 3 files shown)


ftm:
Many thanks for writing the PR up :+1:

PR #538, opened 2020-01-07
fixed SelectMultipleField none_of validator

AbdullahAlajmi:
none_of validator for SelectMultipleField is always true and will never raise a ValidationError.
none_of validator now supports both SelectField and SelectMultipleField.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -594,10 +594,20 @@ class NoneOf(object):
         self.values_formatter = values_formatter
 
     def __call__(self, form, field):
+        if field.type == "SelectMultipleField":
+            if any(e in self.values for e in field.data):
+                message = self.message
+                if message is None:
+                    message = field.gettext('Invalid value, cannot be any of: %(values)s.')
+
+                raise ValidationError(
+                    message % dict(values=self.values_formatter(self.values))
+                )
+
         if field.data in self.values:
             message = self.message
             if message is None:
-                message = field.gettext("Invalid value, can't be any of: %(values)s.")
+                message = field.gettext("Invalid value, cannot be any of: %(values)s.")
 
             raise ValidationError(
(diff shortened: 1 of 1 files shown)


ftm:
The CI is failing due to the issues fixed in #541. If you rebase your branch off of the current `master` you should receive all the necessary changes.

azmeuk:
Also, can you please add some unit tests to this PR?

src/wtforms/validators.py
```diff
@@ -594,10 +594,20 @@ def __init__(self, values, message=None, values_formatter=None):
         self.values_formatter = values_formatter
 
     def __call__(self, form, field):
+        if field.type == "SelectMultipleField":
```
davidism, line 597: This can't be the solution, as it excludes any other fields that have a list of data.

Could maybe do `if isinstance(field.data, list)`. That might have issues with `FieldList`, but that might not matter. Would also have issues if you want to check "none of these specific sets of choices", rather than "doesn't contain any of these choices".

src/wtforms/validators.py
```diff
@@ -594,10 +594,20 @@ def __init__(self, values, message=None, values_formatter=None):
...
+            if any(e in self.values for e in field.data):
+                message = self.message
+                if message is None:
+                    message = field.gettext('Invalid value, cannot be any of: %(values)s.')
+
+                raise ValidationError(
+                    message % dict(values=self.values_formatter(self.values))
+                )
+
         if field.data in self.values:
```
davidism, line 607: If the field is `SelectMultipleField`, this behavior still triggers if the new if block doesn't raise. That would sort of address my point about validating lists vs items, but it doesn't seem like the right design because there's still no way to control *which* type of validation you intended.

azmeuk:
Closing as inactive, and replaced by #807

Issue #529, opened 2019-12-05
Url validator should allow two consecutive dashes in domain


alanhamlett:
This valid url should pass Url validation:

```
https://example--domain.com/file.txt
```

The two consecutive dashes causes validation to fail.

PR #540, opened 2020-02-01
Allow consecutive dashes in hostname

owgreen:
fixes: #529 
This allows consecutive dashed in hostname. Thus, http://foo--bar.dk/ is validated as valid URL.


ftm:
The CI is failing due to a misconfiguration which I have just fixed in #541. If you can base this PR off of the latest `master` then you should get the updated CI config and your tests should pass.

ftm:
Thanks for submitting the PR :+1:

PR #545, opened 2020-03-19
Added switch field.

PanderMusubi:
For description, please see the commit. For examples, see:
* https://getbootstrap.com/docs/4.4/components/forms/#switches
* https://demos.jquerymobile.com/1.4.5/flipswitch/
* https://docs.siimple.xyz/form/switch.html

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -215,6 +215,8 @@ refer to a single input from the form.
 
 .. autoclass:: BooleanField(default field arguments, false_values=None)
 
+.. autoclass:: SwitchField(default field arguments, false_values=None)
+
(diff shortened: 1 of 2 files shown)


ftm:
I don't think this is something we'll be adding into WTForms. As the code in the PR shows a Switch Field would be functionally identical to the pre-existing BooleanField.

PanderMusubi:
Well, there are also different fields for date+time, date and time which have the same underlying datatype too. I hope you would reconsider as this is the most generic way to trigger checkbox of a switch for front-end frameworks.

PanderMusubi:
For those who are interested, it has been implemented in Bootstrap Flask https://github.com/greyli/bootstrap-flask/blame/master/docs/advanced.rst#L134

PR #554, opened 2020-04-17
Drop Python 2, 3.4, 3.5

### LINKED ISSUE #552: End python2 support
I open this ticket to discuss the end of the python2 support. The pallet community [seems to want to end](https://palletsprojects.com/blog/ending-python2-support/) the support, but how is it doable now in wtforms? I suppose that would require a major version bump.

Could wtforms 3 be the version that would drop python 2 support?

azmeuk:
Should fix #552


davidism:
We can also remove 3.5, using `pyupgrade --py36-plus` hits most of it, but f-string conversion doesn't catch all the cases so I usually do some manual search and cleanup.

azmeuk:
Python 3.5 is still supported until 2020-09-13.
Don't you think it worth waiting a few months before dropping py35 support?

davidism:
We decided with Pallets to just drop it early.

davidism:
Since this is in master, 2.3 will still support the old versions, so people will have a relatively updated version to use if they can't upgrade. I'm fine with dropping 3.5 in master.

azmeuk:
Fair enough :)

azmeuk:
Done.

PR #557, opened 2020-04-17
Fixed missing PRs in the changelog

azmeuk:


davidism:
Doesn't have to be now, but we should switch to sphinx-issues at some point, it's a lot less typing.

azmeuk:
I did not know [sphinx-issues](https://github.com/sloria/sphinx-issues), it looks great!

PR #563, opened 2020-04-18
Added a github issue template

azmeuk:
The header message is taken from the flask issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,33 @@
+### Actual Behavior
+
+```python
+# Paste a minimal piece of code that causes the problem. Here is an example:
+
+>>> import wtforms
+>>> class F(wtforms.Form):
+...     foo = wtforms.StringField()
+>>> f = F(foo="bar")
+>>> print(f.foo.shrug)
+Traceback (most recent call last):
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,43 @@
+<!-- **This issue tracker is a tool to address bugs in wtforms itself.
```
davidism, line 1: I'd remove these header comments. GitHub can now show multiple templates, and some of them can be links to other places instead. https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository The comments also show up in the hover cards instead of the first visible content, which makes them less useful. In my experience people ask questions here anyway, so I have essentially this comment as a canned response that I close with.

azmeuk: Removed :)

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,35 @@
+Hello dear WTForms community. I have a problem!
```
davidism, line 1: As nice as this is, it's just going to cause noise most of the time. We can use the more powerful template description metadata to add descriptions of what to use the template for, in a future PR.

PR #568, opened 2020-04-19
Avoid to cache form errors

### LINKED ISSUE #522: Form errors property isn't idempotent
The 'errors' property in BaseForm isn't idempotent. In particular - if you iterate over a forms attributes using e.g. inspect.getmembers - this will cause 'errors' to change from 'None' to dict({}). And so it won't be updated if any fields add/incur errors later in processing.

At a minimum, it should treat 'None' and empty dir the same. But I don't understand why it doesn't re-calc the errors every time?

azmeuk:
Caching form errors may cause some strange behaviors when a form `errors` attribute is accessed during validation, and performance benefits are not obvious.

This should fix #522

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -66,6 +66,7 @@ Unreleased
 -   Choices which name and data are the same do not need to use tuples. (:pr:`526`)
 -   :class:`~validators.URL` validator now allows query parameters in the URL. (:pr:`523`, :pr:`524`).
 -   Updated French and Japanese translations. (:pr:`514`, :pr:`506`)
+-   form.errors is not cached and will update if an error is appended to a field after access. (:pr:`568`)
 
 
 Version 2.2.1
--- a/src/wtforms/form.py
+++ b/src/wtforms/form.py
@@ -31,7 +31,6 @@ class BaseForm(object):
 
         self.meta = meta
         self._prefix = prefix
-        self._errors = None
         self._fields = OrderedDict()
 
         if hasattr(fields, "items"):
@@ -126,7 +125,6 @@ class BaseForm(object):
 
         Returns `True` if no errors occur.
         """
-        self._errors = None
         success = True
         for name, field in iteritems(self._fields):
             if extra_validators is not None and name in extra_validators:
@@ -143,11 +141,7 @@ class BaseForm(object):
 
     @property
     def errors(self):
-        if self._errors is None:
-            self._errors = dict(
-                (name, f.errors) for name, f in iteritems(self._fields) if f.errors
-            )
-        return self._errors
+        return {name: f.errors for name, f in self._fields.items() if f.errors}
 
 
 class FormMeta(type):
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -5,7 +5,7 @@ import pytest
 from wtforms.form import BaseForm, Form
 from wtforms.meta import DefaultMeta
 from wtforms.fields import StringField, IntegerField
-from wtforms.validators import ValidationError
+from wtforms.validators import ValidationError, DataRequired
 from tests.common import DummyPostData
 
 
@@ -229,6 +229,21 @@ class TestForm:
         assert F(DummyPostData()).test.data == "processed"
         assert F(DummyPostData(), test="test").test.data == "processed"
         assert F(DummyPostData({"test": "foo"}), test="test").test.data == "foo"
+
+    def test_errors_access_during_validation(self):
+        class F(Form):
+            foo = StringField(validators=[DataRequired()])
+
+            def validate(self):
+                super(F, self).validate()
+                self.errors
+                self.foo.errors.append("bar")
+                return True
+
+        form = F(foo="whatever")
+        form.validate()
+
+        assert {"foo": ["bar"]} == form.errors
 
 
 class TestMeta:
--- a/docs/forms.rst
+++ b/docs/forms.rst
@@ -47,10 +47,6 @@ The Form class
         A dict containing a list of errors for each field. Empty if the form
         hasn't been validated, or there were no errors.
 
-        Note that this is a lazy property, and will only be generated when you
-        first access it. If you call :meth:`validate` after accessing it, the
-        cached result will be invalidated and regenerated on next access.
-
     .. attribute:: meta
 
         This is an object which contains various configuration options and also


davidism:
I don't think we should remove the caching here. Instead, this should cache a [`ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap) so that updates to the field dicts pass through.

davidism:
Actually nevermind, there's a bigger issue. `form.errors` needs to be writable so you can add form-level errors.

azmeuk:
Interesting. What would be the API then? I can imagine two scenarios:

- The first one is mimicking the way we add errors in Fields, with `myform.errors.append("my error")`. The issue here is that `myform.errors` would behave like a dictionary when reading it, and would behave like a list when writing in it.
- The second one would be just to allow access to the error dictionary. So for instance `myform.errors["my_field"].append("my field error")` would append an error in the field `my_field`. The `None` key could be used to store the form-level errors (not attached to any peculiar field). For instance `myform.errors[None].append("my form error")`.
This second scenario feels more consistent, but adding an form-level error seems tedious. We could then add a `add_error` method that would do this job in both Fields and Forms.

What do you think? I like this second scenario better.

If this API seems OK to you, I will open another PR as I believe it is a different matter than this one. That way we can merge this one?

CHANGES.rst
```diff
@@ -66,6 +66,7 @@ Unreleased
 -   Choices which name and data are the same do not need to use tuples. (:pr:`526`)
 -   :class:`~validators.URL` validator now allows query parameters in the URL. (:pr:`523`, :pr:`524`).
 -   Updated French and Japanese translations. (:pr:`514`, :pr:`506`)
+-   Avoid to cache form errors. (:pr:`568`)
```
davidism, line 69: I'd reword this to be more specific. "`form.errors` is not cached and will update if an error is appended to a field after access."

davidism:
Maybe just make the list of form errors an attribute called `form_errors`, and add it as the `None` key for `errors`. Although I guarantee that would then confuse some users wondering why there's `form.errors` and `form.form_errors`. I wouldn't worry about an `add_error` method for now

It's been a long time since I used form-level errors, maybe it's not actually needed. I'm trying to think of a scenario that I'd use them over just adding the error to a field. I don't think it was ever documented as a feature, so it's probably ok to change the behavior. If someone does report an issue, we can always address it in a bug fix. Given that piece of the documentation you removed that called out when it would be regenerated, I'm thinking form-level errors were just an accident.

I'm still hesitant to remove caching, as users might access `form.errors` multiple times and not know it builds a new dict in a loop every time. Perhaps there's some hybrid we can use where we cache the dict, and add new keys to it if needed each time it's accessed. Not sure if it's really a performance issue, or if my suggestion would be any faster.

I guess overall I'm fine with the PR as-is, we can always address these things later if someone brings them up.

src/wtforms/form.py
```diff
@@ -143,11 +141,7 @@ def data(self):
 
     @property
     def errors(self):
-        if self._errors is None:
-            self._errors = dict(
-                (name, f.errors) for name, f in iteritems(self._fields) if f.errors
-            )
-        return self._errors
+        return dict((name, f.errors) for name, f in iteritems(self._fields) if f.errors)
```
davidism, line 144: This should be a dict comprehension and `iteritems` is no longer needed. `{name: f.errors for name, f in self._fields.items() if f.errors}`

tests/test_form.py
```diff
@@ -230,6 +230,22 @@ class F(Form):
         assert F(DummyPostData(), test="test").test.data == "processed"
         assert F(DummyPostData({"test": "foo"}), test="test").test.data == "foo"
 
+    def test_errors(self):
+        class F(Form):
+            foo = StringField(validators=[DataRequired()])
+
+            def validate(self):
+                if not super(F, self).validate():
+                    return False
```
davidism, line 239: Since this is a test and `return False` will never happen, just make this a direct call to `super().validate()` rather than an if statement, otherwise you have unreachable code. Also don't need the old-style `super` args.

azmeuk: Until #554 is merged there are still python2 tests that may fail if we use the new-style `super()`.

azmeuk:
I fixed your suggestions.

### About perfs

> I'm still hesitant to remove caching, as users might access form.errors multiple times and not know it builds a new dict in a loop every time. Perhaps there's some hybrid we can use where we cache the dict, and add new keys to it if needed each time it's accessed. Not sure if it's really a performance issue, or if my suggestion would be any faster.

I cannot picture a real world case where the caching has significant performance impact:
- forms have a limited amount of fields (probably a few dozen for large ones)
- forms have a limited amount of subforms (FormFields in FormFields in FormFields, but that probably won't go very far)
- I can imagine cases where you wan to access to `form.errors` a few times, but not a lot of times.

Even in the worst case scenario, think that building the dict won't do a large difference. Say a Form containing 100 FormFields themselve 100 FormFields themselve containing 100 Fields, errors for each possible field. So this is about building a dict with 1M entries. Now let us say you access the dict 10 times, here are the profiling results on my computer:

```python
>>> profile.run('for _ in range(10): {x:x for x in range(10^6)}')                                                                                                                                                                                                                      
         14 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 :0(exec)
        1    0.000    0.000    0.000    0.000 :0(setprofile)
       10    0.000    0.000    0.000    0.000 <string>:1(<dictcomp>)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 profile:0(for _ in range(10): {x:x for x in range(10^6)})
        0    0.000             0.000          profile:0(profiler)
```

### About form-level errors

I totally see cases where form-level errors would be pertinent. For instance when you have each individual field that has a correct value, but the whole form do not make sense.
e.g. 4 fields for a store opening time: `morning_opening`, `morning_closing`, `afternoon_opening`, `afternoon_closing`. Each field must have a value that is lesser than the following, when it is not the case, there is not obvious faulty field. In my opinion this is a form-level error.

davidism:
Looks good now.

For your example, I'd add validation for each field that its value is after the previous field, so the error would go on the field that didn't validate. I guess it comes down to a matter of preference. Let's just merge this and leave it for now.

azmeuk:
I will also backport it in branch 2.2

PR #570, opened 2020-04-19
Merge FieldLists postdata with objectdata

### LINKED ISSUE #509: FieldList default data with missing request entries
There is a strange behavior with wtforms 2.2

If you initialize a form with some request data and an object, and you forget a field in the request data, then the form will take the object's value.

```python
>>> import wtforms
>>> from tests.common import DummyPostData
>>> class Form(wtforms.Form):
...     foo = wtforms.StringField()
... 
>>> class Obj:
...     foo = "OK"
... 
>>> Form(DummyPostData({}), Obj()).data
{'foo': 'OK'}
```

Now if you do the same experiment with a `FieldList`, the form will not take the object values.
```python
>>> class Form(wtforms.Form):
...     bars = wtforms.FieldList(wtforms.StringField(), min_entries=3)
... 
>>> class Obj:
...     bars = ["one", "two", "three"]
... 
>>> Form(DummyPostData({'bars-0': 'A'}), Obj()).data
{'bars': ['A', '', '']}
```

Here I would have expected `{'bars': ['A', 'two', 'three']}`.

What do you think? If this is indeed an unwanted behavior, I volunteer to provide a pull request.

azmeuk:
This patch fixes #509, and replaces #510 and #511.

Let us consider this code. It is a simple `Form` containing a `FieldList` with a minimum size of 5:

```python
>>> from wtforms import Form, StringField, FieldList
>>> from tests.common import DummyPostData
...
>>> class F(Form):
...     a = FieldList(StringField(), min_entries=5)
...
>>> a  = F().a
...
>>> pdata = DummyPostData({"a-1": "foo"})
>>> data = ["bar0", "bar1", "bar2"]
>>> a.process(pdata, data)
>>> a.data
["foo", None, None, None, None]
```

From one hand we have object data with a list containing 3 elements. From the other hand we have some post data for that `FieldList` but providing only 1 index, and not even the first one.

What should happen here?
- Should the post data overwrite the object data? That would mean here creating a new list, with only one element, then fill the list with empty values to reach the minimum size: `["foo", None, None, None, None]` This is how WTForms currently behaves.
- Should the post data update the object data? That would mean take the object data, replace indices provided by the form data, then fill the list with empty values to reach the minimum size: `["bar0", "foo", "bar2", None, None]`.

I am in favor of merging data, this is what this PR does.
1. [explicit is better than implicit](https://www.python.org/dev/peps/pep-0020/) so:
  - object data explicitly passed to the form should be used
  - if you do not send data for one index, then it should not be modified
2. If you handle a `ListField` of `FileField` and have files already present in you object data, you just cannot send them back in the post data just to keep them in the list.
3. it seems consistent with the general behavior for forms: If you send post data but forget one field, the forgotten fields are not resetted: 
```python
>>> class G(Form):
...     a = StringField()
...     b = StringField()
...
>>> class Obj:
...     a = "A"
...     b = "B"
...
>>> F(DummyPostData({"a": "foo"}), Obj()).data
{'a': 'foo', 'b': 'B'}
```

If think this behavior should be the default, but if you think it is too controversial, maybe we can make it an option?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -66,6 +66,7 @@ Unreleased
 -   Choices which name and data are the same do not need to use tuples. (:pr:`526`)
 -   :class:`~validators.URL` validator now allows query parameters in the URL. (:pr:`523`, :pr:`524`).
 -   Updated French and Japanese translations. (:pr:`514`, :pr:`506`)
+-   Merge FieldLists postdata with objectdata (:pr:`560`)
 
 
 Version 2.2.1
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -1047,19 +1047,20 @@ class FieldList(Field):
             if self.max_entries:
                 indices = indices[: self.max_entries]
 
-            idata = iter(data)
-            for index in indices:
-                try:
-                    obj_data = next(idata)
-                except StopIteration:
-                    obj_data = unset_value
-                self._add_entry(formdata, obj_data, index=index)
+            max_index = max(max(indices), len(data))
+            for index in range(max_index + 1):
+                if index < len(data) or index in indices:
+                    obj_data = data[index] if index < len(data) else unset_value
+                    self._add_entry(formdata, obj_data, index=index)
         else:
             for obj_data in data:
                 self._add_entry(formdata, obj_data)
 
         while len(self.entries) < self.min_entries:
-            self._add_entry(formdata)
+            if len(self.entries) < len(data):
+                self._add_entry(formdata, data[len(self.entries)])
+            else:
+                self._add_entry(formdata)
 
     def _extract_indices(self, prefix, formdata):
         """
--- a/tests/test_fields.py
+++ b/tests/test_fields.py
@@ -928,14 +928,14 @@ class TestFieldList:
         assert not form.validate()
 
         form = F(pdata, a=data)
-        assert form.a.data == ["bleh", "yarg", "", "mmm"]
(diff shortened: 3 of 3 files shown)


davidism:
My comment from the first PR:

> You're passing in form data with one value, then manually processing some other data and expecting the two to be merged, which doesn't seem correct.

I'll try to explain a bit better.

`data` is only *default* data, this is true in general of fields. Users regularly mix this up and are confused when some fields no longer have data after they process `formdata` that doesn't include that field. Despite this confusion, this is the intended behavior and the only one that makes sense, otherwise you couldn't clear a field with existing data.

So despite `FieldList` being a more complex field, this change still seems to be break the intended behavior. For example, if you initially have three fields filled out, there's no way to cut that down to one field with this change.

davidism:
Your comment on `FileField` is true, but this isn't the solution. I think that's a case where we need better documentation. "If you're using a `FieldList` of `FileField`, existing files won't be re-sent, so you'll want to update the existing data with any new posted data manually."

That problem exists with plain `FileFields` too, a field won't have any existing data. File inputs are just weird in general. It would be interesting to explore a more complex `ClearableFileField` like what Django provides. This would actually be three fields, a hidden input with a marker for existing data, a file input for new data, and a checkbox for deleting the data.

azmeuk:
I thought a bit about this topic, and I think the main issue I have is with `FileField`, so indeed, it is another subject.

I guess we could document `ListField` a bit more on that matter. I saw several tickets where people seem to not always find this behavior intuitive.

PR #575, opened 2020-04-21
Version 2.3.0

azmeuk:


davidism:
Uploaded 2.3.0 to [PyPI](https://pypi.org/project/WTForms/2.3.0/) (and [Test PyPI](https://test.pypi.org/project/WTForms/2.3.0/)). Docs with the new theme are at https://wtforms.readthedocs.io/en/2.3.x/. I've adjusted the docs URLs so that we can push docs fixes to the 2.3.x and they'll be updated (in the past they were stuck at tagged versions). Now to merge 2.3.x into master.

PR #582, opened 2020-04-22
Bugfix, tests and documentation for SelectField choices shortcut

### LINKED ISSUE #572: Documentation and tests for SelectField choices shortcuts
#526 allows `SelectField` to have choices that are not tuples. However, there is no documentation and [no test](https://coveralls.io/builds/30176772/source?filename=src/wtforms/fields/core.py#L524) about this feature.

azmeuk:
Fixes #572 and [this bug](https://github.com/wtforms/wtforms/pull/526#issuecomment-617675939) introduced with #526

### DIFF
--- a/wtforms/fields/core.py
+++ b/wtforms/fields/core.py
@@ -476,7 +476,9 @@ class SelectField(SelectFieldBase):
         self.validate_choice = validate_choice
 
     def iter_choices(self):
-        if isinstance(self.choices[0], (list, tuple)):
+        if not self.choices:
+            choices = []
+        elif isinstance(self.choices[0], (list, tuple)):
             choices = self.choices
         else:
             choices = zip(self.choices, self.choices)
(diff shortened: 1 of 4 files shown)


wtforms/fields/core.py
```diff
@@ -476,7 +476,7 @@ def __init__(self, label=None, validators=None, coerce=text_type, choices=None,
         self.validate_choice = validate_choice
 
     def iter_choices(self):
-        if isinstance(self.choices[0], (list, tuple)):
+        if self.choices and isinstance(self.choices[0], (list, tuple)):
```
spookey, line 479: I would not do it that way...

If someone forgot to explicitly set the choices (leaving them at ``None``) then the else clause is triggered on rendering. ``zip(None, None)`` then raises ``TypeError: zip argument must support iteration``.

spookey: On the other hand - I am very impressed by your quick reaction. Thank you very much!

azmeuk: Good catch.
Eh, staying home has at least some benefits :)

PR #584, opened 2020-04-22
fixed extra_validators parameter

azmeuk:
[BaseForm.validate](https://wtforms.readthedocs.io/en/2.3.x/forms/#wtforms.form.BaseForm.validate) method could take a `extra_validators` parameter, but [Form.validate](https://wtforms.readthedocs.io/en/2.3.x/forms/#wtforms.form.Form.validate) could not.

If this is ok I would also like to backport it to branch 2.3.x

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@ Unreleased
 -   Restored ``HTMLString`` and ``escape_html`` as aliases for
     MarkupSafe functions. Their use shows a ``DeprecationWarning``.
     :issue:`581`, :pr:`583`
+-   ``Form.validate`` takes an ``extra_validators`` parameter, mapping
+    field names to lists of extra validator functions. This matches
+    ``BaseForm.validate``. :pr:`584`
 
 
 Version 2.3.0
--- a/tests/form.py
+++ b/tests/form.py
@@ -151,6 +151,23 @@ class FormTest(TestCase):
         form = self.F()
         self.assertEqual(form.validate(), False)
 
+    def test_validate_with_extra(self):
+        class F2(self.F):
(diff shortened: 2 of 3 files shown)


davidism:
Rebased this to 2.3.x, added a changelog.

I copy `extra_validators` with `dict` if it was passed in. This is to avoid the case where a dict is defined separately and passed to the function. If `validate` were called multiple times in that case, `validate_fieldname` would get appended to the same list multiple times. `BaseForm.validate` and `Field.validate` don't modify the value, so they don't need to be changed.

azmeuk:
I am not sure that copy is needed here. I think just having `None` as the default argument prevents the mutable default argument trap.

```python
>>> d = {}
>>> def foobar(d=None): 
...     _d = d or {} 
...     _d.setdefault("foo", []).append("bar") 
...     return _d
>>> foobar(d)
{'foo': 'bar'}
>>> foobar(d)
{'foo': 'bar'}
```
However that would have been the case with `d={}` instead of `d=None`.
```python
>>> def foobar(d={}): 
...     d.setdefault("foo", []).append("bar") 
...     return d
>>> foobar(d)
{'foo': ['bar']}
>>> foobar(d)
{'foo': ['bar', 'bar']}
```

davidism:
Your first test code doesn't seem to match up, you show it producing a string value instead of a list.

azmeuk:
:) bad copy paste, now the snippet is accurate

davidism:
Ah, it's failing because your `d` dict is empty. If you do `d = {"existing": ["value"]}` you'll see the issue.

azmeuk:
oooK now I see :)

PR #591, opened 2020-04-24
pytest for 2.3.x

azmeuk:
Because even fixing small bugs in painful without pytest


davidism:
Need to switch this over to GitHub actions too.

This needs to match *exactly* the way it looks in master, minus any new tests master added, otherwise it's going to be really hard to merge upwards.

azmeuk:
Ok, I will port the `self.assert*` syntax to `assert`.
We might need to `black` everything too.

davidism:
Maybe we don't need to do this? I don't anticipate many more changes to 2.3.x, but `python tests/runtests.py` isn't too bad. I can try to set up GitHub actions, but I've just been running tox locally to confirm the changes in 2.3.1.

davidism:
3.x should be the next release. We can close very old/broad issues related to refactoring, @ftm and I had a previous discussion in Discord about not doing such a radical refactor.

Issue #606, opened 2020-05-19
SelectMultipleField does not implement validate_choice parameter


Nyaanbinary:
### Actual Behavior

```python
>>> import wtforms
>>> class F(wtforms.Form):
...     single = wtforms.SelectField(choices=[], validate_choice=False)
...     multiple = wtforms.SelectMultipleField(choices=[], validate_choice=False)
>>> f = F()
>>> f.single.process_data(1)
>>> f.multiple.process_data([1])
>>> f.validate()
False
>>> f.single.errors
[]
>>> f.multiple.errors
["'1' is not a valid choice for this field"]
```

### Expected Behavior

```python
>>> import wtforms
>>> class F(wtforms.Form):
...     single = wtforms.SelectField(choices=[], validate_choice=False)
...     multiple = wtforms.SelectMultipleField(choices=[], validate_choice=False)
>>> f = F()
>>> f.single.process_data(1)
>>> f.multiple.process_data([1])
>>> f.validate()
True
>>> f.single.errors
[]
>>> f.multiple.errors
[]
```
The ```pre_validate``` function in the ```SelectMultipleField``` class overwrites the functionality for ```validate_choice``` present in the ```pre_validate``` function of its parent class, ```SelectField```. Thus, the ```validate_choice``` parameter in ```SelectMultipleField``` is completely disregarded when the form's ```validate``` function is called, resulting in a wrongful validation error.
I believe this functionality should be implemented in order to add additional functionality consistency to the API.

### Environment

* Python version: 3.7.6
* wtforms version: 2.3.1

azmeuk:
Hi. Thank you for your bug report.
As you have a pretty good insight on how to fix this issue, would you consider submitting a patch?

PR #607, opened 2020-05-20
Added validate_choice to SelectMultipleField

Nyaanbinary:
Added implementation for the ```validate_choice``` parameter to ```SelectMultipleField``` based on its implementation in SelectField.

Original issue page where I described this problem:
https://github.com/wtforms/wtforms/issues/606

This edit is adding a simple if statement to the ```pre_validate``` function of ```SelectMultipleField```, based on the ```pre_validate``` function of ```SelectField```.

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -576,14 +576,15 @@ class SelectMultipleField(SelectField):
             )
 
     def pre_validate(self, form):
-        if self.data:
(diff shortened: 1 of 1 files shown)


azmeuk:
Thank you very much for you patch.
Please run *black* to fix the failing style check.
Also, please provide some unit tests so we are sure a future commit won't break your fix. I suppose you can adapt some of the `SelectField` tests.

PR #609, opened 2020-05-24
allow uuid object on uuid validations

### LINKED ISSUE #549: UUID validator fails if field.data is a UUID object instead of str
The current implementation of `UUID` validator at https://github.com/wtforms/wtforms/blob/master/src/wtforms/validators.py#L520 assumes that field.data is of type `str`. However, `if field.data` is already an instance of `uuid.UUID` the validator fails with the following exception: `AttributeError: 'UUID' object has no attribute 'replace'`.

I've found this issue working with PostgreSQL, SQLAlchemy and using Flask-Admin with a field declared as:

    from sqlalchemy.dialects.postgresql import UUID

    # ... More code here
    uuid = db.Column(
        UUID(as_uuid=True), nullable=False, unique=True,
        default=uuid4, server_default=sa.text('uuid_generate_v4()'))

Note the use of `as_uuid=True` as said at https://github.com/flask-admin/flask-admin/issues/1444.

My proposal is to change the line `uuid.UUID(field.data)` by `uuid.UUID(str(field.data))`, this way it should work in both cases.

Meanwhile, I've done the following workaround implementing a new UUID validator overriding the `__call__()` method as follows:

    import uuid

    from wtforms.validators import UUID

    class SecureUUID(UUID):
            if isinstance(field.data, uuid.UUID):
                field.data = str(field.data)
            super().__call__(form, field)

owgreen:
Fixes #549 

Allow uuid object on validations of UUID.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -521,7 +521,7 @@ class UUID:
         if message is None:
             message = field.gettext("Invalid UUID.")
         try:
-            uuid.UUID(field.data)
+            uuid.UUID(str(field.data))
         except ValueError:
             raise ValidationError(message)
 
--- a/tests/test_validators.py
+++ b/tests/test_validators.py
@@ -1,5 +1,6 @@
 import decimal
 import re
+import uuid
 
 import pytest
 
@@ -300,7 +301,11 @@ def test_bad_mac_raises(mac_addr_val, dummy_form, dummy_field):
 
 @pytest.mark.parametrize(
     "uuid_val",
-    ["2bc1c94f-0deb-43e9-92a1-4775189ec9f8", "2bc1c94f0deb43e992a14775189ec9f8"],
+    [
+        "2bc1c94f-0deb-43e9-92a1-4775189ec9f8",
+        "2bc1c94f0deb43e992a14775189ec9f8",
+        uuid.uuid4(),
+    ],
 )
 def test_valid_uuid_passes(uuid_val, dummy_form, dummy_field):
     """


whb07:
what is this PR doing here again? Its directly pulling some UUID object and then you're passing it directly into a form for validation? Can you explain the workflow that gets accomplished by this please?

thijs-obj:
Coincidentally, this PR happens to fix another problem: if the data is `None` then the UUID validator gives a `TypeError` instead of ` ValueError`. and as a result `form.validate()` will raise an exception instead of returning `False`

azmeuk:
@whb07 I think the workflow is explained in #549. Basically, in the [flask-admin](https://github.com/flask-admin/flask-admin) project, a uuid model attribute ends up generating a `StringField` with a `UUID` validator, and gives a `UUID` directly to the validator instead of a string.

HTML form data are always `str` but usually fields cast them into convenient types, so validators deals with casted data instead of raw `str`. For instance `IntegerField` casts the `str` value into a `int`, then `NumberRange` only expects to handle numeric types.

There is no UUID field so the UUID validator expects to handle strings. When the field data is not string, this causes the validator to crash:

```python
>>> import wtforms
>>> import uuid
>>> class F(wtforms.Form):
...     foo = wtforms.StringField(validators=[wtforms.validators.UUID()])
>>> class MyObject:
...     foo = uuid.uuid4()
>>> f = F(obj=MyObject())
>>> f.foo.data
UUID('33e7a9f7-19f5-4eae-aa76-051a2c28815f')
>>> f.validate()
Traceback (most recent call last):
  ...
  File "/home/azmeuk/.local/lib/python3.8/site-packages/wtforms/validators.py", line 487, in __call__
    uuid.UUID(field.data)
  File "/usr/lib/python3.8/uuid.py", line 166, in __init__
    hex = hex.replace('urn:', '').replace('uuid:', '')
AttributeError: 'UUID' object has no attribute 'replace'

'UUID' object has no attribute 'replace'
```

However, I don't know if this example is good practice. Can we validate a Form without HTML form data? I suppose nothing prevents that, even if the usual way to validate a form is `if request.POST and form.validate():` @davidism what do you think?

azmeuk:
Another usecase : passing an UUID without casting to the `default` parameter.
```python
>>> import wtforms
>>> import uuid
>>> class F(wtforms.Form):
...     foo = wtforms.StringField(
...         validators=[wtforms.validators.UUID()],
...         default=uuid.uuid4(),
...     )
>>> F().validate()
Traceback (most recent call last):
...
'UUID' object has no attribute 'replace'
```

azmeuk:
Similar behavior with `DateTimeFields` and `datetime.datetime` as default value.

```python
>>> import wtforms
>>> import datetime
>>> class F(wtforms.Form):
...     foo = wtforms.DateTimeField(
...         default=datetime.datetime.now(),
...     )
>>> F().validate()
Traceback (most recent call last):
...
TypeError: strptime() argument 1 must be str, not datetime.datetime
```

davidism:
I thought there was a broader issue about this, where we discussed whether fields and validators should accept strings, or also accept their types. Maybe I'm getting confused with Click, which has a similar issue with command line strings and parameter types.

davidism:
It also seems like we'd want UUID to be a field, not a validator, since I'd expect to work with `UUID` objects in Python, not strings that happen to be valid UUIDs.

azmeuk:
I have been thinking about this. In my opinion `StringField` should cast any input into string, so the validator could safely assume that the field data will be a string. I suppose that would require implementing a `StringField.process_data` method to do the cast.

src/wtforms/validators.py
```diff
@@ -521,7 +521,7 @@ def __call__(self, form, field):
         if message is None:
             message = field.gettext("Invalid UUID.")
         try:
-            uuid.UUID(field.data)
+            uuid.UUID(str(field.data))
         except ValueError:
             raise ValidationError(message)
```
TurnrDev, line 526: It might be best to check if the type is uuid.UUID instead of blindly casting to str

```suggestion
        if isinstance(field.data, uuid.UUID):
            return

        try:
            uuid.UUID(field.data)
        except ValueError:
            raise ValidationError(message)
```

tests/test_validators.py
```diff
@@ -300,7 +301,11 @@ def test_bad_mac_raises(mac_addr_val, dummy_form, dummy_field):
 
 @pytest.mark.parametrize(
     "uuid_val",
-    ["2bc1c94f-0deb-43e9-92a1-4775189ec9f8", "2bc1c94f0deb43e992a14775189ec9f8"],
+    [
+        "2bc1c94f-0deb-43e9-92a1-4775189ec9f8",
+        "2bc1c94f0deb43e992a14775189ec9f8",
+        uuid.uuid4(),
```
TurnrDev, line 307:
```suggestion
        uuid.UUID("2bc1c94f-0deb-43e9-92a1-4775189ec9f8"),
```

RNG in tests is bad!

azmeuk:
Closing as superseded by #769, and anyways there is [a larger design question](https://github.com/wtforms/wtforms/issues/806) to be discussed first.

PR #610, opened 2020-05-25
Update validators.py

awx1:
FIxes a bug in DataRequired where the data for time 00:00:00 is not recognized as a correct time but rather as None.

Hardcodes the case where this is unrecognized and recognizes it as a correct time rather than None.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -265,7 +265,10 @@ class DataRequired:
 
     def __call__(self, form, field):
         if not field.data or isinstance(field.data, str) and not field.data.strip():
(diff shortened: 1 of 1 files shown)


davidism:
`bool(time(0, 0, 0))` returns `True` on all supported Python version. It looks like you're seeing Python 2.7 behavior. Python 2 is no longer supported.

When submitting a pull request, please be sure to include a test that demonstrates the issue and ensure all tests pass.

awx1:
My bad. I didn't mean to make a bad request. I was wondering, is there any
fix to this bug in Python 2.7?

PR #628, opened 2020-06-23
pass RadioField (SelectFieldBase) render_kw dict in _Option class

dizcza:
> Describe the issue you are attempting to fix.

RadioField ignores `render_rw` dict arg. I noticed that when I was trying to pass `render_rw=dict(disabled=True)` as an argument in RadioField constructor.

Populating `_Option` with RadioField's `.render_kw` attribute solves the issue.

However, there is an issue with your testing setup that does not allow me to pass `render_rw` while creating a radio field to make a regression test.

-_-

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 3.0.0
 
 Unreleased
 
+-   Fixed :class:`~fields.RadioField` validators. :issue:`477` :pr:`615`
+
 Version 3.0.0a1
 ---------------
 


dizcza:
Ah, I see. WTForms devs are very _responsive_ towards others fixing their bugs.

Very nice, good job, developers! :+1:

tests/test_fields.py
```diff
@@ -678,6 +678,17 @@ class F(Form):
             "</ul>"
         )
 
+    def test_render_kw_preserved(self):
+        F = make_form(
+            a=RadioField(
+                choices=[(True, "yes"), (False, "no")],
+                render_rw=dict(disabled=True)
```
azmeuk, line 685: There is a typo here `render_rw` -> `render_kw`.

azmeuk:
With the typo fixed, the test pass without your fix.

tests/test_fields.py
dizcza: Thank you! Fixed.

dizcza:
@azmeuk  There is also an annoying style check. Do I need to follow it blindly?

tests/test_fields.py
```diff
@@ -678,6 +678,17 @@ class F(Form):
...
 
+    def test_render_kw_preserved(self):
+        F = make_form(
+            a=RadioField(
+                choices=[(True, "yes"), (False, "no")],
+                render_kw=dict(disabled=True)
+            )
+        )
+        form = F()
+        assert 'disabled' in form.a()
```
azmeuk, line 689: This assertion is true even without your patch. Please write a test that fails without your patch, and succeeds with your patch.

azmeuk:
Yes. But you can simply run `black` on the files you have edited.

azmeuk:
This ticket seems related to #490

tests/test_fields.py
dizcza: I wrote a test to check for exact match - I don't know how to check it otherwise.

dizcza:
Done. Let me know if I need to do anything else.

dizcza:
Hope that you squash the changes during the merge.

azmeuk:
Wow. I am not sure how this issue get closed, I just rebased your branch against wtforms master.

davidism:
I'm not sure how, but this has happened to me a few times too. You might be able to use `git reflog` to find the branch before you rebased it, then try again.

PR #637, opened 2020-07-20
New validator: NetworkAddress

fportantier:
It's a copy of IPAddress, but adapted for ipaddress.ip_network()

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -16,9 +16,9 @@ __all__ = (
     "EqualTo",
     "equal_to",
     "IPAddress",
+    "ip_address",
     "NetworkAddress",
     "network_address",
-    "ip_address",
     "InputRequired",
     "input_required",
     "Length",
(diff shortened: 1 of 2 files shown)


src/wtforms/validators.py
```diff
@@ -466,6 +468,70 @@ def check_ipv6(cls, value):
         return True
 
 
+class NetworkAddress:
+    """
+    Validates a Network address. Requires ipaddress package to be instaled for
```
whb07, line 473: typo in "installed"

whb07: Is the 'ipaddress' package a dependency?

davidism:
I'm not comfortable with how much of a copy this is. It basically creates two identical implementations that we have to keep in sync for bug fixes, docs, etc. Is it possible to rework this to build on the existing validator, maybe as a subclass?

That said, and especially if refactoring is not possible, I'm not sure we really need this in WTForms. It's much less common than a plain `IPAddress`, and projects or extension packages can maintain whatever validators they need.

PR #639, opened 2020-07-29
Backported #598

azmeuk:
Backport of #598 for the v2 branch as suggested in https://github.com/lepture/flask-wtf/issues/417.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 .. currentmodule:: wtforms
 
 
+Unreleased
+----------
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 .. currentmodule:: wtforms
 
 
+Unreleased
```
davidism, line 4: This should be "Version 2.3.2", with "Unreleased" under it. Can fix that when you make the release PR.

PR #640, opened 2020-07-29
Bumped to 2.3.2

azmeuk:
v2.3.2


azmeuk:
@davidism I will let you handle it from here as I do not have access to pypi for wtforms

davidism:
You should have access on both PyPI and Test PyPI now.

azmeuk:
Ok thanks!

Issue #646, opened 2020-09-01
Validation fails on required boolean fields when they have valid data


drakkhen:
When a required boolean field is set to False, validation fails.  I would expect validation to fail if the field's value was NOT set, or something other than a boolean value.  But a required boolean field should validate successfully if it's set to True or False.

### Actual Behavior

```python
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import wtforms
>>> wtforms.__version__
'2.3.3'
>>> class F(wtforms.Form):
...      foo = wtforms.BooleanField('Has Foo', [wtforms.validators.InputRequired()])
...
>>> f = F(foo=False)
>>> f.validate()
False

```

### Expected Behavior

```python
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import wtforms
>>> wtforms.__version__
'2.3.3'
>>> class F(wtforms.Form):
...      foo = wtforms.BooleanField('Has Foo', [wtforms.validators.InputRequired()])
...
>>> f = F(foo=False)
>>> f.validate()
True
```

### Environment

* Python version:  3.8.5
* wtforms version:  2.3.3

azmeuk:
Duplicate of #402

PR #647, opened 2020-09-01
#646 - Allow required BooleanField instances to have a value of False

drakkhen:
The previous logic in InputRequired.__call__ was only testing to see if
field.raw_data was "truthy", but False is a valid value for a BoolenField.

This change tests BooleanField and non-BooleanField instances differently,
and will require a required BooleanField to have a value of True or False.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -2,6 +2,7 @@ import ipaddress
 import math
 import re
 import uuid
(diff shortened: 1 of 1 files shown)


azmeuk:
Related (again) to #402. We should find a whole policy for wtforms, not just a specific fix for `BooleanField`.

PR #648, opened 2020-09-07
Fix typo in German translation of email address

jugmac00:
In Germany you have to write "E-Mail-Adresse" - "Email" is no valid
German.

also see
https://www.duden.de/rechtschreibung/E_Mail_Adresse

modified:   src/wtforms/locale/de/LC_MESSAGES/wtforms.po


azmeuk:
Thank you.

azmeuk:
@jugmac00 This is out in v 3.0.0a now.

PR #649, opened 2020-09-07
correction in class DateTimeField() def _value() to hande data of typ…

georgerififi:
correction in class DateTimeField() def _value() to handle data of type datetime.timedelta()


davidism:
`timedelta` is not intended to work with this field, and `timedelta.strfdelta` does not appear to exist. If you need a field for timedeltas, you can write your own `Field` subclass to do what you need for your project.

PR #650, opened 2020-09-07
 correction in class DateTimeField() def _value() to hande data of type datetime.timedelta

georgerififi:
correction in class DateTimeField() def _value() to hande data of type datetime.timedelta


davidism:
Duplicate of #649

PR #654, opened 2020-09-17
Add min and max html tag attributes to the RangeInput widget

sergehauri:
Add `min` and `max` html tag attributes to the RangeInput widget equivalent to the NumberInput widget.

### DIFF
--- a/src/wtforms/widgets/core.py
+++ b/src/wtforms/widgets/core.py
@@ -491,12 +491,18 @@ class RangeInput(Input):
     input_type = "range"
     validation_attrs = ["required", "max", "min", "step"]
 
-    def __init__(self, step=None):
+    def __init__(self, step=None, min=None, max=None):
         self.step = step
+        self.min = min
+        self.max = max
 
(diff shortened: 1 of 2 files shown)


azmeuk:
Actually I am not sure that `NumberInput` should have `min` and `max` attributes in the first place. I suppose the HTML `min` and `max` parameter should only be populated with the `NumberRange` validator.
@davidism do you have a thought?

davidism:
Not sure, I could see a case either way. `RangeInput` already has the `step` parameter, but that's because `step` isn't really related to validation (although it could be). Widgets don't do validation, so the params would be purely cosmetic, which might be confusing to users.

sergehauri:
Thanks for your feedback. I now realize I was using v2.3.1, where adding the NumberRange validator didn't add the HTML attributes. I should have checked first if it has already been solved in v3 - my mistake, sorry.
Having the validator add the HTML attributes is a much better solution, so I agree the min and max, as well as the step attributes don't really make sense in the widget.

Thanks for watching out for the library!

PR #655, opened 2020-09-26
class TimeDeltaField

georgerififi:
new class handles datetime.timedelta data


davidism:
Duplicate of #649. Thank you, but I do not want to add custom handling of timedeltas to the core library. Feel free to use this field in your own code.

PR #657, opened 2020-10-04
Re-enable coverage

azmeuk:
@davidism in #578 you said:

> Coverage makes the tests slower, produces artifacts locally, and we're not using the reports right now.

As `coverage` is not in the tox envlist, it is not launched with `tox` and you should explicitly run `tox -e coverage` to run it. Is it acceptable to you?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -66,7 +66,6 @@ filterwarnings =
 
 [coverage:run]
 branch = true
-parallel = true
 source =
     wtforms
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,6 +65,7 @@ filterwarnings =
     error
 
 [coverage:run]
+branch = true
 parallel = true
 source =
     wtforms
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -65,14 +65,14 @@ filterwarnings =
     error
 
 [coverage:run]
-branch = true
+parallel = true
```
davidism, line 68: You'll want to keep `branch` and not `parallel`. Branch coverage is important, and parallel mode should be enabled at the call in the tox config (and isn't needed in the config you've written since there's only one env).

davidism:
Yeah, adding a config like this is fine.

setup.cfg
azmeuk: For some reasons I do not understand, without `parallel = true` the coverage report paths are `.tox/py38/.../src/wtforms` instead of `src/wtforms/`

https://hynek.me/articles/testing-packaging/

davidism: Ah, this is one of the reasons I got frustrated with coverage. :smile:

PR #659, opened 2020-10-05
Reject float input in IntegerFields

ocaballeror:
IntegerFields happily take float input and coerce it to int, losing the decimal part without informing the user in any way.

By converting input to string every time we ensure that the value is a valid python int every time, no matter its original format.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,12 @@ Unreleased
     bugfix. :issue:`606` :pr:`642`
 
 
+Unreleased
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for looking at this. After thinking about it a bit, I don't think this should change. I understand that the current behavior could be seen as allowing data that isn't valid according to the type, but it matches the behavior of Python's `int`: if the input is a string, it must look like an int, but if it's numeric in general it will be coerced.

PR #663, opened 2020-10-07
Allow to use regexp on multiplefilefield

### LINKED ISSUE #442: Regexp validator fails on MultipleFileField
Hi Guys,

Regexp validator fails on multiple file field - it validates on field.data which in this situation is a list.

(I wanted to clone the repo and try out new Regexp test case with list type field.data value, but I failed at running pytest on non-installed package and picking up whole validators tests...)

Cheers

azmeuk:
Fixes #442


azmeuk:
Hm. I will do it another way.

PR #664, opened 2020-10-07
DateTimeField (and others) supports multiple formats

### LINKED ISSUE #175: Multiple datefield format
It would be nice if datefield and datetimefield could accept multiple date format. From country to country, the date format changes, so this would be very helpful.

azmeuk:
Fixes #175 
`DateTimeField`, `DateField`, `TimeField` and `MonthField` can be given lists of format instead of a single format to match the input.
If several formats are available, only the first one is used for HTML output.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,9 @@ Unreleased
     bugfix. :issue:`606` :pr:`642`
 -   Fixed SelectMultipleField validation when using choices list shortcut.
     :issue:`612` :pr:`661`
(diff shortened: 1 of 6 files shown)


davidism:
While the implementation seems fine, I'd generally discourage this. In my opinion it's better to be exact about what is accepted, and communicate that to the user. For example, add a help text or placeholder showing the format, use the modern HTML widgets which already use and validate a standard format, or add JavaScript for validation for browsers that don't support one of the widgets.

If you expect the user to be entering values free form, it would be better to parse that with `dateutil.parse` rather than trying to guess which formats they might use (which will also differ based on locale).

PR #665, opened 2020-10-07
Python 3.9 support

azmeuk:
Python 3.9 tox and github actions tests.


azmeuk:
It seems 3.9 is not available in github actions yet.

tirkarthi:
Relevant actions issue : https://github.com/actions/setup-python/issues/148

PR #675, opened 2020-12-28
Always call setattr in FormField.populate

antonstakhouski:
I'm writing a flask website that communicates with an API gateway using json. I wrote a some ORM models using python dataclasses that convert json to objects and vice versa and hide all data manipulation stuff.

The problem is that i could receive a nested json: `{"foo": 42, "nested": {"bar": "baz"}}`. And i map it to something like
```python
@dataclass
class B(BaseModel):
    bar: str = None

@dataclass
class A(BaseModel):
    foo: str = None
    nested: dict = None
    
    @property
    def nested_(self):
          return B.from_dict(self.nested)
    
    @property
    def nested_(self, value):
        self.nested = value.to_dict
```

And form is:
```python
class NestedForm(FlaskForm):
   bar = StringField('Bar')

class MyForm(FlaskForm):
   foo = StringField('Foo')
   nested_ = FormField(NestedForm, 'Nested')
````

In routes I have code like this:
```python
def update_a():
    form = MyForm()
    if form.validate_on_submit:
         obj = A()
         form.populate_obj(obj)
         obj.save()
```
`FormField` will not modify `obj.nested` (it won't write temporary `candidate` object to `obj` attribute).

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -1032,9 +1032,9 @@ class FormField(Field):
                     " the provided obj or input data/defaults"
                 )
             candidate = self._obj
-            setattr(obj, name, candidate)
 
         self.form.populate_obj(candidate)
+        setattr(obj, name, candidate)
 
     def __iter__(self):
         return iter(self.form)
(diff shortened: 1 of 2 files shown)


azmeuk:
I played a bit with your patch and I could not find anything wrong with this. 
@davidism any thougths on this?

davidism:
Seems fine, go ahead and merge it when ready.

PR #676, opened 2021-01-04
New validators

Ennkua:
Added new validators

DigitRequired:
This validator ensures that the incoming data must contain a specified minimum number of digits.
:param min:
The minimum required digit in the string.
min default value is 1 digit at least.
:param message:
Error message to raise in case of a validation error.
----------------------------------------------------------------------------------------------------------------------------
UppercaseRequired:
This validator ensures that the incoming data must contain a specified minimum number of uppercase letters.
:param min:
The minimum required uppercase letter in the string.
min default value is 1 uppercase letter at least.
:param message:
Error message to raise in case of a validation error.
----------------------------------------------------------------------------------------------------------------------------
LowercaseRequired:
This validator ensures that the incoming data must contain a specified minimum number of lowercase letters.
:param min:
The minimum required lowercase letter in the string.
min default value is 1 lowercase letter at least.
:param message:
Error message to raise in case of a validation error.
----------------------------------------------------------------------------------------------------------------------------
SymbolRequired:
This validator ensures that the incoming data must contain a specified minimum number of symbols.
:param min:
The minimum required symbol in the string.
min default value is 1 symbol at least.
:param message:
Error message to raise in case of a validation error.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,7 +47,8 @@ Released 2020-11-23
     bugfix. :issue:`606` :pr:`642`
 -   Fixed SelectMultipleField validation when using choices list shortcut.
     :issue:`612` :pr:`661`
-
+-   Added ``DigitRequired``, ``UppercaseRequired``, ``LowercaseRequired``
+    and ``SymbolRequired`` validators.
 
 Version 2.3.3
 -------------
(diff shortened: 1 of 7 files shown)


davidism:
Thanks, but I don't plan to add these to the core library. You're welcome to turn these into your own PyPI package.

Issue #658, opened 2020-10-05
NumberRange step validation


azmeuk:
As mentioned in #654, it could be interesting to add a `step` argument to the `NumberRange` validator. This way it could validate some discrete values. A value `x` should be accepted if `min <= x <= max` and `x - min % step == 0`.

For instance `NumberRange(min=10, max=20, step=3)` would only accept values in `[10, 13, 16, 19]`

davidism:
I think this would only work reliably for ints though, in which case the following should also work:

```python
if int(value) == value and value not in range(min, max, step):
    raise ValidationError(...)

... # original min/max only checks

**Comment by azmeuk** [comment]
I could imagine cases with `FloadField`, `DateField` or `MonthField` too.

**Comment by davidism** [comment]
I don't think `step` can work (or at least be accurate) for float, but I'm not 100% sure. Dates aren't numbers, that would probably require a different validator that used something like `dateutil.rrule`.

**Comment by whb07** [comment]
It essentially ends up being the issue that everyone has with floats and precision/truncation, right?

I could imagine something naive being
```python
step = "0.01"
_, step_val = step.split(".")
value = "1.123"
_, decimal = value.split(".")
if len(decimal) >  len(step_val):
    raise Error
```

After that initial check, its a modulo check for the step sizing validation.

azmeuk:
6 years later, I cannot think of a real world use case for this. Let us close this and reopen it if someone comes up with an example.

PR #679, opened 2021-02-08
FloatField use NumberInput by default

### LINKED ISSUE #677: FloatField should use a NumberInput by default
`FloatField` uses `TextInput` but should instead use a `NumberInput` with step `any` by default, somewhat like `DecimalField` does.

### Actual Behavior

```python
>>> import wtforms
>>> class F(wtforms.Form):
...     foo = wtforms.FloatField()
>>> F().foo()
Markup('<input id="foo" name="foo" type="text" value="">')
```

### Expected Behavior

```python
>>> import wtforms
>>> class F(wtforms.Form):
...     foo = wtforms.FloatField()
>>> F().foo()
Markup('<input id="foo" name="foo" type="number" value="" step="any">')
```

### Environment

* Python version: 3.9
* wtforms version: 3.0.0a1

Demetriex:
FloatField should use NumberInput 

Fix #677


whb07:
Hi @Demetriex , this looks good! A couple things to add perhaps is:

Add some tests verifying the behavior:

- When the "step" property is not defined as "any" or a specific float, then the field doesn't pass validation for that specific float. The HTML input itself when calling `input.checkValidity()` wont report `true` unless its "any" or a specific float up to the defined precision, i.e - step="0.01" validates for "1.10" but not for "1.234"

I also left some comments on the changes for you to take a look at.

src/wtforms/fields/core.py
```diff
@@ -818,7 +818,7 @@ class FloatField(Field):
     is ignored and will not be accepted as a value.
```
whb07, line 818: The documentation for this class needs to be updated

tests/fields/test_float.py
```diff
@@ -13,9 +13,11 @@ def test_float_field():
     form = F(DummyPostData(a=["v"], b=["-15.0"]))
```
whb07, line 13: Need to test the "step" values association to the float input data. So a step of "any" is good with "1.234" but a step of "0.1" should not validate a "1.234" input.

whb07 (changes requested):
Looks good. Just a couple more checks and validations to follow the spec.

tests/fields/test_float.py
Demetriex: thanks for your input, it looks like this related to this issue #658.

PR #683, opened 2021-02-19
fix: when field value is False, whink field is missing(BooleanField)

susnmos:

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -287,7 +287,7 @@ class DataRequired:
         self.field_flags = {"required": True}
 
     def __call__(self, form, field):
-        if field.data and (not isinstance(field.data, str) or field.data.strip()):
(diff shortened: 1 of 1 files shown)


whb07:
hi @susnmos , can you expand on what you're trying to accomplish here? The "truthy" check for Python can be thought of as:

```python

# false empty containers
empty_dict = {}
empty_list = []
empty_tuple = ()
# falsey things
value = None
value = False
```
And so on. So if thats true then that means that we can perform the following check in one time:
```python
data = [] # <---- can also be None here
if not data:
    raise ValueError
```

But a None and an empty list is not equal! If you're trying to show a better way to boil down the boolean logic of that entire line, then please submit a revised PR.

azmeuk:
I will close for now. @susnmos we can reopen when you describe what this patch does and why.

PR #684, opened 2021-02-24
SelectMultipleField: Check that valuelist is not None in process_formdata before modifying field.data

sarambulo:
The bugg: When you create a Form using formdata and key-word arguments, the SelectMultipleField would be assign and empty array, instead of the value in formdata (or the value in the key-word argument as a backup).

The fix: Check that the valuelist argument in SelectMultipleField.process_formdata() is not None before changing the field's data attribute. This check is already been made for SelectField.process_formdata(). That is why the problem only arises with SelectMultipleField.

There are a couple of new assert statements in 'test_selectmultiple.py' to check that the problem has been fixed.

PD: There was an error in 'test_selectmultiple.py:40'. The value of 'b' should be equal to the default value define in line 18. But 'b' had a value of '[]' because of the bug in cuestion. That is why the test used to pass, but was now giving an error.

### DIFF
--- a/src/wtforms/fields/core.py
+++ b/src/wtforms/fields/core.py
@@ -603,6 +603,9 @@ class SelectMultipleField(SelectField):
             self.data = None
 
     def process_formdata(self, valuelist):
+        if not valuelist:
+            return
+
         try:
             self.data = list(self.coerce(x) for x in valuelist)
         except ValueError:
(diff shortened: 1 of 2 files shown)


davidism:
This is incorrect, keyword arguments are the same as defaults, they're only used if no form data was passed.

PR #703, opened 2021-08-03
add compatibility for removing leading-zeros in strftime/strptime

briancappello:
The datetime format specifier for `datetime.strftime` allows to remove leading zeros from outputted data. However, attempting to use the same format string with `datetime.strptime` raises with `ValueError: '-' is a bad directive in format '%-m/%-d/%Y %-I:%M %p %z'`. This PR fixes the issue.


azmeuk:

PR #704, opened 2021-08-07
Field Prefix gets overriden by Form Prefix

olliebreeden:
If you set a field _prefix value, it always gets overriden by the form _prefix as self.kwargs contains the field _prefix.

I think these should be combined so it's possible to have both form and field prefix.


davidism:
Field _prefix has an underscore because it's not intended as a public API. It only makes sense to give a form a prefix, not an individual field.

Issue #662, opened 2020-10-06
Field data overhaul


azmeuk:
I think there are some inconsistencies in the way data is handled by fields:

## How things works now
Here is a small reminder on how inputs are handled in forms and fields.

### input
There are several ways to *put* data into a field:

- with request form data  (`Form.process()` `formdata` argument); 
- with a placeholder object data (`Form.process()` `obj` argument);
- with a placeholder data dict (`Form.process()` `data` argument);
- with a field default value (`Field` `default` argument).

### process
Then in `Field.process()`:

- `Field.process_formdata()` processes the form data. The input in always a list of strings. The form data is stored in `Field.raw_data`, then it may be coerced in a convenient python type, and is finally stored the result in `Field.data`;
- `Field.process_data()` processes the object data value, or the data dict value, or the field default value (in that order). The input can have any type. The form data is stored in `Field.object_data`, then it may be coerced in a convenient python type, and is finally stored the result in `Field.data`.
- In the end `filters` are successively applied, and may transform `Field.data`.

`process_formdata()` and `process_data()` may find errors and add them into `Field.process_errors`

### validate
Validation consists in:

- `Field.pre_validate()`;
- inline and extra validators;
- `Field.post_validate()`.

Generally `pre_validate()`, `post_validate()` and the validators handle `Field.data`, that is python data. In addition to the process errors `validate()` adds the validation errors into `Field.errors`.

### output

`Field._value()` returns a value to display for the widgets. Generally it transforms `Field.data` in a *htmlized* form, but sometimes it takes shortcuts by returning the field formdata input (in `Field.raw_data`):
https://github.com/wtforms/wtforms/blob/9cc7c7f4d32c8e69c67460e8b8b2aa5ed411da2e/src/wtforms/fields/core.py#L707-L712

### both `process_formdata()` and `process_data()` are executed

If the field has some `formdata`, then `process_formdata()` will overwrite `self.data` if it has been previously set by `process_data()`. It seems useless to execute `process_data()` at all then.

https://github.com/wtforms/wtforms/blob/9cc7c7f4d32c8e69c67460e8b8b2aa5ed411da2e/src/wtforms/fields/core.py#L342-L356

- [ ] I suggest not executing `process_data()` if `formdata is not None`.

### validation is always attempted
As discussed in #61, #360, #435, even if `Field.process_data()` or `Field.process_formdata()` raise a `ValueError`, and then `Field.validate()` is called, then validation on that field will be attempted.

[...]

PR #716, opened 2021-12-05
Amend `InputRequired` validator to support multi-input fields

dalepotter:
Small change to the `InputRequired` validator, so that it examines the entire contents of `field.raw_data` for validation logic (rather than just the first element). This is helpful when the HTML representation of a `Field` renders more than one `<input>`- e.g. using a widget.

For example, a WTForms implementation of the [Gov.UK Design System Date Input component](https://design-system.service.gov.uk/components/date-input/) renders three HTML `<input>` elements. Prior to this fix, missing input in the first (i.e. Day, or `raw_data[0]`) input triggers a `ValidationError`, even if data was entered in the second (i.e. Month, or `raw_data[1]`) and/or third (i.e. Year, or `raw_data[2]`) inputs.

Commit checklist:

- [x] add tests that fail without the patch - see 9797106d
- [x] ensure all tests pass with ``pytest``
- N/A add documentation to the relevant docstrings or pages
- [x] add ``versionadded`` or ``versionchanged`` directives to relevant docstrings - see 7a87422
- [x] add a changelog entry if this patch changes code - see 7a87422

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Version x.x.x
 -------------
 Unreleased
 
+- :class:`~validators.InputRequired` Validator interrogates data on entire
+  contents of field `raw_data`. :pr:`719`
+
 Version 3.0.0
 -------------
 
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -308,6 +308,9 @@ class InputRequired:
     looks at the post-coercion data.
 
     Sets the `required` attribute on widgets.
+
+    .. versionchanged:: 3.0,1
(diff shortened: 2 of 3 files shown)


azmeuk:
Hi. Thank you for your patch. I am very interested in the subject of wtforms being able to manage several input. This may be related to #662. I think I have to take some time to think of all the implications, so that would be when I am a little less busy. But I do not forget.

azmeuk:
I am closing this because at the moment wtforms does not support several inputs with the same name. This might come in the future, but that would require a big overhaul.
Related to #799

PR #726, opened 2022-01-08
Update documentation on Custom widgets

nattster:
1. Fix wrong id used in the example
2. Add example on how to use the custom widget function in the form


azmeuk:
Thank you.

PR #732, opened 2022-02-09
Added missing Korean translations.

josueisonfire:
I have added the missing Korean translations, though I have to ask, the translation for the string, 
```"Choices cannot be None."``` 
is a bit tricky, due to `None` being a Python programming term, and there is no direct translation for it (the same applies for null). Korean programmers will most likely understand if I refer to it as `None`, so I left it that way.
If  ```"Choices cannot be None. "``` is asking the user that an empty field is invalid, the transaltion
```"선택 값이 없을 수 없습니다."``` (meaning 'the choice cannot be empty') is more appropriate.


azmeuk:
Hi. Thank you for you contribution. I cannot really help to review korean. Is this patch OK to merge or do you prefer to be sure of your translation first?

josueisonfire:
Yes, I think the translation is good as it is.
Sorry for the confusion, and thanks!

PR #737, opened 2022-04-08
Improve documentation on quantization of `DecimalField`

gentooboontoo:
* Specify how many decimal places are used by default to quantize value if left unset
* Clarify that decimal places should be intentionaly  set to `None` for not quantizing value


PanderMusubi:
@azmeuk can you also review this one? Thanks.

azmeuk:
Thank you!

Issue #381, opened 2018-04-10
SelectField option_widget argument not used


nathan-steyer:
The `option_widget` parameter to SelectField is not used by the built-in `Select` widget to render the choices as `option` elements. It appears to use its `render_option` class method regardless.

ftm:
If nothing uses it, I think the `option_widget` parameter should be removed as unless I'm mistaken there's nothing else you can use other than `<option>` which is the default.

davidism:
It's definitely used elsewhere. See `RadioField` for an example:

```python
class RadioField(SelectField):
    widget = widgets.ListWidget(prefix_label=False)
    option_widget = widgets.RadioInput()
```

`SelectField` could use some cleanup, it's a bit circuitous right now.

azmeuk:
Could reproduce with:

```python
>>> from markupsafe import Markup
>>> from wtforms import Form, SelectField, Choice
>>> class TaggedOption:
...     def __call__(self, field, **kwargs):
...         return Markup(
...             f'<option data-custom="yes" value="{field.choice.value}">'
...             f'!{field.choice.label}!</option>'
...         )
...         
>>> class F(Form):
...     color = SelectField(option_widget=TaggedOption(), choices=[Choice("r", "Red"), Choice("g", "Green")])
...     
>>> form = F(color="g")
>>> form.color()
Markup('<select id="color" name="color"><option value="r">Red</option><option selected value="g">Green</option></select>')
```

azmeuk:
This should be done at the same time than #715

PR #739, opened 2022-04-28
`SelectField` choice refactoring

### LINKED ISSUE #692: SelectField - can choices tuples be expanded to allow for tooltips
At present the choices list should contain tuples:
`[ ("option-value", "display-text"), (..., ...), ... ]`

Can it be expanded to optionally allow a **tool-tip** as 3rd tuple element which is then rendered as a `title` attribute in the <option> element:
`("option-value", "display-text", "tool-tip")` --> `<option value="optional-value" title="tool-tip">display-text</option>`

azmeuk:
Fixes #692. Related to #381.

`SelectField` `choice` argument can take a list of 3 items tuples, the last one being a dict that will be rendered as `<option>` HTML parameter.

### DIFF
--- a/src/wtforms/fields/choices.py
+++ b/src/wtforms/fields/choices.py
@@ -49,8 +49,10 @@ class SelectFieldBase(Field):
             _form=None,
             _meta=self.meta,
         )
-        for i, (value, label, checked) in enumerate(self.iter_choices()):
-            opt = self._Option(label=label, id="%s-%d" % (self.id, i), **opts)
+        for i, (value, label, checked, render_kw) in enumerate(self.iter_choices()):
+            opt = self._Option(
+                label=label, id="%s-%d" % (self.id, i), **opts, **render_kw
+            )
             opt.process(None, value)
             opt.checked = checked
             yield opt
@@ -112,8 +114,9 @@ class SelectField(SelectFieldBase):
         else:
             _choices = zip(choices, choices)
 
-        for value, label in _choices:
-            yield (value, label, self.coerce(value) == self.data)
+        for value, label, *other_args in _choices:
+            render_kw = other_args[0] if len(other_args) else {}
+            yield (value, label, self.coerce(value) == self.data, render_kw)
 
     def process_data(self, value):
         try:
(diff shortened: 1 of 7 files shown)


davidism:
Seems good to me. Instead of passing longer tuples, you might want to consider a `Choice` dataclass with the fields, which might make typing easier later.

azmeuk:
Do you mean we should break the compatibility and expect the `choices` parameter to be a collection of `SelectChoice` instead of a collection of tuples?
I suppose this would make the code more explicit and easier to read though.

davidism:
Don't add a 3-tuple data type. Instead, only support this new data in the new dataclass. If you get a string, or a 2-tuple, convert it to the dataclass.

Since `choices` can be set after init, I'd do this conversion in the render step for now (and maybe in init as well). I'd probably keep the string shortcut, but show a deprecation warning for 2-tuples like "Passing (value, label) is deprecated, pass SelectChoice(...) instead.".

azmeuk:
Sorry for the noise with the force pushes, that should be good now.

I refactored the `SelectField` by adding a `Choice` dataclass. I deprecated passing choices by tuples or dicts. The `Choice` dataclass carry the `optgroup` and `render_kw`.
This was a good idea @davidism I did not realized of much this would simplify the `SelectField` behavior :+1: 

I could use another review :pray:

src/wtforms/fields/choices.py
```diff
@@ -1,16 +1,63 @@
...
+        if isinstance(input, str):
+            return Choice(value=input, optgroup=optgroup)
+
+        if isinstance(input, tuple):
+            warnings.warn(
+                "Passing SelectField choices as tuples is deprecated and will be "
+                "removed in wtforms 3.3. Please use Choice instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
ThiefMaster, line 57: Please reconsider this. It will be a significant amount of busy-work for larger projects to adapt to this, without any real benefit. Just generating the Choice instance from the tuple is cheap and doesn't seem to have any drawbacks.

image

Yeah... I really do not want to spend time updating those almost 50 places in my app.

coleifer:
This breaks a few libraries, including `wtf-peewee` which passes two-tuples.

PR #741, opened 2022-05-04
Allow passing of test_environment flag to email_validator.

Lukas0907:
email_validator rejects most special-use reserved domain names such as `test` as of 1.2. A new `test_environment` option has been added to allow `@*.test` domains. This PR allows passing of the option from wtforms to email_validator.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Version 3.0.2
 Unreleased
 
 -   Delayed import of ``email_validator``. :issue:`727`
+-   Pass ``test_environment`` flag to ``email_validator``.
 
 Version 3.0.1
 -------------
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ install_requires = MarkupSafe
 where = src
 
 [options.extras_require]
-email = email_validator
+email = email_validator >= 1.2.0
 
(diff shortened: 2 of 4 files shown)


azmeuk:
Thank you!

davidism:
All in this list except https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Special-Use_Domains except `.onion` and `.invalid` should probably be accepted. `.example`, `.local`, `.localhost`, `.test`. Note that `.test` and `.local` are specifically reserved for DNS, not email, although that's unlikely to matter.

davidism:
Also, it's not clear how you're meant to set this flag during tests only.

davidism:
@azmeuk this should probably be reverted if the use case is not clarified

azmeuk:
Done. Sorry for this, at first sight it looked ok to me.

Lukas0907:
> All in this list https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains#Special-Use_Domains except .onion and .invalid should probably be accepted.

The email validation is still done by the email validator plugin. The flag just modifies its validation behavior.

> Also, it's not clear how you're meant to set this flag during tests only 

Well, an environment variable could be used for this but I do agree it's a bit tedious.

PR #759, opened 2022-11-13
for Regexp validator set html input tag 'pattern' attribute

lefty01:
for Regexp validator set html input tag 'pattern' attribute  (used if _novalidate_ not set on a form)

Similar to the Length validator that sets the html input attributes minlength and maxlength I found it useful to have this for regex patterns as well.
One advantage of using the browser is to get the message translated already (even tough here it is a bit generic, since it only says input needs to match the requested format and does not tell about the format itself)

added some basic test ... not sure if that is accepted this way?


azmeuk:
Thank you for you contribution. 
I think an additional test that actually generate some HTML would be good to have, maybe in the fashion of https://github.com/pallets-eco/wtforms/blob/7418bac10d689c8ca428e66d465317d9fcd7b215/tests/fields/test_html5.py#L148

Also, can you add a CHANGELOG entry?

For the record, the related doc : https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern

tests/validators/test_regexp.py
```diff
@@ -48,10 +48,29 @@ def test_regex_raises(re_pattern, re_flags, test_v, dummy_form, dummy_field):
...
 def test_regexp_message(dummy_form, dummy_field, grab_error_message):
     """
     Regexp validator should return given message
     """
     validator = regexp("^a", message="foo")
     dummy_field.data = "f"
     assert grab_error_message(validator, dummy_form, dummy_field) == "foo"
+
+
+def test_regexp_pattern_html(dummy_form, dummy_field):
```
azmeuk, line 69: I am not sure of the point of this test. I suppose the assertion could go on `test_regex_passes`.

lefty01:
... need to check again .. I was not that sure about proper testing here. will take a  closer look Thx!

azmeuk:
From the MDN documentation.

> The pattern attribute is an attribute of the [text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text), [tel](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel), [email](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email), [url](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url), [password](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password), and [search](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search) input types.

You would probably want to add the `pattern` flag to the concerned widgets too.

azmeuk:
@lefty01 are you interested in finishing the PR or would you like to let someone else take it over?

ThiefMaster:
IMHO this should be configurable. You might not always want to send the pattern to the client, or you may want to use regex features not supported in the browser.

lefty01:
> @lefty01 are you interested in finishing the PR or would you like to let someone else take it over?

well, interested yes of course but right now it's not on top of my list ;)
so I'm quite happy there's now further discussion ongoing and of course do not mind at all if you/someone takes over the item so we can close it at some point.
... been a while and its bit hard to follow up anyways ;)

dougthor42:
FYI: In my experience with HTML `pattern`, there are two caveats:
1. I've found that the HTML pattern string needs to be slightly different than a standard python re.pattern string. IME the difference is typcially related to escaping.
    + Example: Python accepts `r"[-\w]*"`, but HTML `pattern` must be `r"[\-\w]*"`
3. It doesn't seem possible to "clean" a verbose regex back to a bare pattern string, so any patterns that _are_ passed to HTML `pattern` can't be verbose. See https://stackoverflow.com/q/35640871/1354930

azmeuk:
I added a `html_parameter` param to the regex validator.
- If False, the default, no html pattern is added.
- If True, the python pattern is used
- If str, a custom pattern is used
- This can also be a callable that produce a str or a bool, if people want to try applying conversion methods.

Issue #450, opened 2018-10-01
DateTimeLocalField Parse Error


attack68:
The HTML5 DateTimeLocalField inherits from core.DateTimeField but I am not certain these express the same value formats.

In particular looking at the [Mozilla Docs datetime-local](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) the instruction is clear: seconds are not supported and the string format is `"%Y-%m%-%dT%H:%M"`. Indeed this format is what I get when testing development in Chrome.

It fails at line 742 of fields/core.py (wtforms version 2.2.1) since the format is not correct:

    def process_formdata(self, valuelist):
        if valuelist:
            date_str = ' '.join(valuelist)
            try:
                self.data = datetime.datetime.strptime(date_str, self.format)  #line: 742
                # date_str = "2018-10-10T12:12"
                # self.format = '%Y-%m-%d %H:%M:%S'
                # -> returns: ValueError: time data '2018-10-10T12:12' does not match format '%Y-%m-%dT%H:%M:%S'
            except ValueError:
                self.data = None
                raise ValueError(self.gettext('Not a valid datetime value'))

Of course the downstream form validation then fails since my required input's data are set to `None`.

attack68:
Rather than defining my own this was of course very simple and something I overlooked.

Just specifiy the `format` directly according to standards:

    datetime = DateTimeLocalField('Input DateTimeLocal', format='%Y-%m-%dT%H:%M', validators=[InputRequired()])

PR #761, opened 2022-11-22
Add shorter format datetime-local defaults

russellfinlay:
### Describe the issue you are attempting to fix

This issue was previously raised in #450.

The HTML standard for datetime-local inputs set the default *step* attribute to 60, meaning that the lowest denomination of input can be expressed in minutes, not seconds.

https://html.spec.whatwg.org/#local-date-and-time-state-(type=datetime-local)

> The step attribute is expressed in seconds. The step scale factor is 1000 (which converts the seconds to milliseconds, as used in the other algorithms). The default step is 60 seconds.

In such cases, it seems that common browser behaviour is to then send the input value, without any seconds, on form submission, in the format -
```
2022-11-02T12:23
```

The current default of WTForms is to validate against two possible date formats, both that require seconds -

```python
def __init__(self, *args, **kwargs):
        kwargs.setdefault("format", ["%Y-%m-%d %H:%M:%S", "%Y-%m-%dT%H:%M:%S"])
```

If seconds are not found, the validation will fail -
```
Not a valid datetime value.
```

### Current fixes

This issue can be fixed by adding the correct format to a DateTimeLocalField() instance -

```python
example_date_time = DateTimeLocalField('Example datetime', format="%Y-%m-%dT%H:%M")
```

or by change input step size to allow seconds (in HTML)**
```html
<input type="datetime-local" step="1">
```

### Proposal

Very simple PR to bring WTForms in line with the default behaviour of some browsers, by adding two formats (without seconds) to the default formats of DateTimeLocalField.

```python
    def __init__(self, *args, **kwargs):
        kwargs.setdefault("format", ["%Y-%m-%d %H:%M:%S", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M", "%Y-%m-%dT%H:%M"])
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 -   Delayed import of ``email_validator``. :issue:`727`
 -   Python 3.11 support :pr:`763`
+-   Added shorter format to :class:`~fields.DateTimeLocalField`
+    defaults :pr:`761`
 
 Version 3.0.1
 -------------
--- a/src/wtforms/fields/datetime.py
+++ b/src/wtforms/fields/datetime.py
@@ -123,5 +123,5 @@ class DateTimeLocalField(DateTimeField):
     widget = widgets.DateTimeLocalInput()
 
     def __init__(self, *args, **kwargs):
-        kwargs.setdefault("format", ["%Y-%m-%d %H:%M:%S", "%Y-%m-%dT%H:%M:%S"])
+        kwargs.setdefault("format", ["%Y-%m-%d %H:%M:%S", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M", "%Y-%m-%dT%H:%M"])
         super().__init__(*args, **kwargs)
--- a/src/wtforms/fields/datetime.py
+++ b/src/wtforms/fields/datetime.py
@@ -123,5 +123,13 @@ class DateTimeLocalField(DateTimeField):
     widget = widgets.DateTimeLocalInput()
 
     def __init__(self, *args, **kwargs):
-        kwargs.setdefault("format", ["%Y-%m-%d %H:%M:%S", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M", "%Y-%m-%dT%H:%M"])
+        kwargs.setdefault(
+            "format",
+            [
+                "%Y-%m-%d %H:%M:%S",
+                "%Y-%m-%dT%H:%M:%S",
+                "%Y-%m-%d %H:%M",
+                "%Y-%m-%dT%H:%M",
+            ],
+        )
         super().__init__(*args, **kwargs)


azmeuk:
Thank you for your contribution.
Indeed I suppose this field should work out-of-the-box with common browsers.
Can you add a CHANGELOG entry and fix the *style* tests?

azmeuk:
Thank you. The *style* test is still failing though.

russellfinlay:
That should make the tox style testing pass now. It seems that black wanted to re-format that line.

I've also seen that two tests are failing from test_datetime.py. Would you like me to look at those before merge?

azmeuk:
The GHA tests look OK. Are you seeing issues with `test_datetime.py` when you locally run the tests? What error message do you see?

russellfinlay:
Two tests failing when invoking `pytest`, one in `tests\fields\test_datetime.py` and the other in `tests\fields\test_datetimelocal.py`.

I've just re-run the tests without the changes from this pull request and I'm still seeing the same failures, so I imagine that it doesn't have anything to do with this commit.


`tests\fields\test_datetime.py`
```python
__________________________________________________________________ test_basic ___________________________________________________________________ 

    def test_basic():
        d = datetime(2008, 5, 5, 4, 30, 0, 0)
        # Basic test with both inputs
        form = F(
            DummyPostData(
                a=["2008-05-05", "04:30:00"], b=["2008-05-05 04:30"], c=["5/5/2008 4:30"]
            )
        )
        assert form.a.data == d
        assert (
            form.a()
            == """<input id="a" name="a" type="datetime" value="2008-05-05 04:30:00">"""
        )
        assert form.b.data == d
        assert (
            form.b()
            == """<input id="b" name="b" type="datetime" value="2008-05-05 04:30">"""
        )
        assert form.c.data == d
        assert (
            form.c() == """<input id="c" name="c" type="datetime" value="5/5/2008 4:30">"""
        )
        assert form.validate()

        # Test with a missing input
        form = F(DummyPostData(a=["2008-05-05"]))
        assert not form.validate()
        assert form.a.errors[0] == "Not a valid datetime value."

        form = F(a=d, b=d, c=d)
        assert form.validate()
        assert form.a._value() == "2008-05-05 04:30:00"
        assert form.b._value() == "2008-05-05 04:30"
>       assert form.c._value() == "5/5/2008 4:30"

tests\fields\test_datetime.py:52:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <wtforms.fields.datetime.DateTimeField object at 0x0000018C16BD84D0>

    def _value(self):
        if self.raw_data:
            return " ".join(self.raw_data)
>       return self.data and self.data.strftime(self.format[0]) or ""
E       ValueError: Invalid format string

src\wtforms\fields\datetime.py:36: ValueError
```

`tests\fields\test_datetimelocal.py`
```python
__________________________________________________________________ test_basic ___________________________________________________________________ 

    def test_basic():
        d = datetime(2008, 5, 5, 4, 30, 0, 0)
        # Basic test with both inputs
        form = F(
            DummyPostData(
                a=["2008-05-05", "04:30:00"], b=["2008-05-05 04:30"], c=["5/5/2008 4:30"]
            )
        )
        assert form.a.data == d
        assert (
            form.a()
            == '<input id="a" name="a" type="datetime-local" value="2008-05-05 04:30:00">'
        )
        assert form.b.data == d
        assert (
            form.b()
            == '<input id="b" name="b" type="datetime-local" value="2008-05-05 04:30">'
        )
        assert form.c.data == d
        assert (
            form.c()
            == '<input id="c" name="c" type="datetime-local" value="5/5/2008 4:30">'       
        )
        assert form.validate()

        assert form.validate()
        assert form.a._value() == "2008-05-05 04:30:00"
        assert form.b._value() == "2008-05-05 04:30"
>       assert form.c._value() == "5/5/2008 4:30"

tests\fields\test_datetimelocal.py:53:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <wtforms.fields.datetime.DateTimeLocalField object at 0x0000018C16A64250>    

    def _value(self):        if self.raw_data:
            return " ".join(self.raw_data)
>       return self.data and self.data.strftime(self.format[0]) or ""
E       ValueError: Invalid format string

src\wtforms\fields\datetime.py:36: ValueError
```

PR #788, opened 2023-05-25
docs: wtforms/validators: tweak `InputRequired` docs

woodruffw:
Hello!

This is just a small docstring change: no behavioral changes have been made. As such, I haven't added tests or a changelog entry.

This change clarifies the behavior of the `InputRequired` validator by making its behavior around raw vs. coerced data slightly more explicitly documented: it adds an additional sentence ephasizing that "form-input data" means the `formdata` parameter going into a `Form`.

As a justification for this clarification: we're doing a bit of refactoring on Warehouse, including replacing usages of `DataRequired` with `InputRequired`. As part of that, we've observed that a large number of unit tests incorrectly assumed that `Foo(x=1)` has the same validation behavior as `Foo(dict(x=1))`, which it doesn't for fields whose validators look at the field's `raw_data`.

Please let me know if there's anything else I can do here!

cc @jleightcap

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -300,9 +300,9 @@ class InputRequired:
 
     Note there is a distinction between this and DataRequired in that
     InputRequired looks that form-input data was provided, and DataRequired
-    looks at the post-coercion data. This means that fields must be passed
-    in through a `Form`'s `formdata` parameter in order to pass this
-    validator; fields passed in as keyword arguments will fail validation.
+    looks at the post-coercion data. This means that this validator only checks
+    whether non-empty data was sent, not whether non-empty data was coerced
+    from that data. Initially populated data is not considered sent.
 
     Sets the `required` attribute on widgets.
     """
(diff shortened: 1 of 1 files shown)


woodruffw:
xref https://github.com/pypi/warehouse/pull/13696

davidism:
Data passed in as keyword arguments or with the `obj` or `data` parameters is default populated data, not data that should be validated directly. You populate the initial data, render the template, then validate the data that is sent back. This is true of all validators and all initial data, not the `InputRequired` validator on keyword arguments.

I think this change is obscuring that distinction. What the sentence is saying is that both validators are only validating sent data, but `Input` is checking that non-empty data was sent, while `Data` is checking that non-empty data was coerced from that sent data.

woodruffw:
Sorry for the delay!

I see your point -- my goal was to emphasize the existing behavior, but I don't think I've made it clearer.

How does something like this sound instead?

> This means that this validator only checks whether non-empty data was sent, not whether non-empty data was coerced from that data. Initially populated data is not considered sent.

That's still a little clunky, though. Please feel free to close if you think it isn't a net improvement in that form either 🙂

davidism:
Yeah, that sounds good.

woodruffw:
Cool, thanks. Updated!

PR #790, opened 2023-06-18
Adding A Not Equal Validator

khezam:
This PR adds a `NotEqualTo` validator.

None.

This PR adds a `NotEqualTo` validator.


davidism:
Thanks, but I don't think this needs to be added to WTForms core. In general we encourage projects to write small validators for their specific needs, rather than adding more generic ones to core beyond what's already there. You'll note that there's not less than, greater than or equal to, etc. validators either.

khezam:
@davidism :)  

Ah okay, I unserstand that. I will close it then.

Thank you!

PR #797, opened 2023-07-22
implement a readonly and a disabled validator

### LINKED ISSUE #792: HTML `readonly` attribute support
Currently wtforms does not handle the [HTML readonly attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly). One can make a field read-only at rendering with the `render_kw` attribute.

On several projects I have been programmatically setting readonliness to fields, depending on the context, the user permissions etc. In the end I find myself writing the same tests and the same security guards again and again. Before processing the form I need to curate the readonly fields to protect against users sending modified values anyways. I also need to write the associated unit tests. This feels like boilerplate that could be avoided if wtforms would take care of this.

The simpler solution I can think of is to define a `readonly` flag for Fields. The presence of the `readonly` flag would add the HTML `readonly` attribute, and make the `formdata` values to be ignored by [Field.process](https://github.com/wtforms/wtforms/blob/0524993ee57a957c8444df2b090c00464ef9bb2a/src/wtforms/fields/core.py#L292). I did not have played with the code to verify my hypothesis though.

We could have a similar thing for the `disabled` attribute.

@davidism any thoughts?

Related discussions on SO:
https://stackoverflow.com/a/16576294/2180332
https://stackoverflow.com/a/9849156/2180332

Related ticket: #217

azmeuk:
Fixes #792 

This adds two validators [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly) and [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled).

The validators adds the corresponding flags on the widgets on which this is authorized, according to MDN.

The `readonly` validator raises a `ValidationError` if the formdata is different than the object data for a given form. The `disabled` validator raises a `ValidationError` if the formdata is different than `None`.

Readonly fields are sent by the browser, and disabled ones are not. Either ways those validators prevents the user from cheating and submitting different values that the ones they have been presented.

If someone wants a readonly field but does not want to enforce the immutability, they still can use render_kw.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -33,6 +33,10 @@ __all__ = (
     "UUID",
     "ValidationError",
     "StopValidation",
+    "readonly",
+    "ReadOnly",
+    "disabled",
+    "Disabled",
 )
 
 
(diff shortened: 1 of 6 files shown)


azmeuk:
I should look how this behaves  for `FieldList` and `FormField`.

src/wtforms/validators.py
```diff
@@ -676,6 +680,40 @@ def __call__(self, hostname):
...
+        if field.data != field.object_data:
+            raise ValidationError(field.gettext("This field cannot be edited"))
+
+
+class Disabled:
+    """
+    Set a field disabled.
+
+    Validation fails if the form data is different than the
+    field object data, or if unset, from the field default data.
```
ThiefMaster, line 704: That's not correct

azmeuk: Fixed. Now this is `Validation fails if the form data has any value.`

PR #798, opened 2023-07-23
move from setup.cfg to pyproject.toml+hatch

azmeuk:
This should carry the same metadata as before, however the babel catalog compilation on development installation or sdist and wheel building is lost. I am not sure how to do this with pyproject.toml


pyproject.toml
```diff
@@ -0,0 +1,98 @@
+[project]
+name = "WTForms"
+description = "Form validation and rendering for Python web development."
+readme = "README.rst"
+license = {file = "LICENSE.rst"}
+maintainers = [{name = "WTForms", email = "davidism@gmail.com"}]
```
davidism, line 6: You can remove my email from this.

pyproject.toml
```diff
@@ -0,0 +1,98 @@
...
+"Issue Tracker" = "https://github.com/wtforms/wtforms/issues"
+Chat = "https://discord.gg/pallets"
+
+[project.optional-dependencies]
+email = ["email_validator"]
+
+[build-system]
+requires = [
+    "flit_core<4",
+    "Babel>=2.6.0"
```
davidism, line 35: Make sure we're still building the translations before building the dist files.

azmeuk:
This should be better now. compiled catalogs are compiled when the package is built and then correctly included

PR #800, opened 2023-07-31
add allow_ip flag availability from URL validator

Fausterkun:
URL validator validate url and ip (without possibility to exclude ip). That because in URL.call method at initialization HostnameValidation allow_ip flag hard-coded to True.

Added argument allow_ip to URL.init method.


azmeuk:
Thank you.

Issue #811, opened 2023-10-11
Incorrect docstring?


Nicholas-Read:
Barely a bug at all, but the docstring describes iter_choices [here ](https://github.com/wtforms/wtforms/blob/master/src/wtforms/widgets/core.py#L335) is out of sync with the code [just below](https://github.com/wtforms/wtforms/blob/master/src/wtforms/widgets/core.py#L365), which unpacks a tuple of length 4!

Also on the core class [here](https://github.com/wtforms/wtforms/blob/master/src/wtforms/fields/choices.py#L33)!

Thanks for the great library!

aminalaee:
Yeah I think this wasn't a backwards-compatible change and already broke some code: https://github.com/aminalaee/sqladmin/pull/644

davidism:
Sorry, I thought you were saying that multiple docs were out of sync, you actually were saying that the docs were out of sync with the code.

aminalaee:
Yeah I think that's the initial discussion, but it also breaks the code. I guess it's related to: https://github.com/wtforms/wtforms/pull/739

azmeuk:
I have fixed the examples and added a message in the changelog about the compatibility break.

azmeuk:
About your point on old unmaintained projects breaking, and people having to spend time to update their code, well there is a tradeoff to be made:
People ask for bugfixes and features in wtforms, that leads to refactoring, that sometimes leads to compatibility breakages. Keeping the old API puts the maintenance burden on the wtforms maintainers instead of the maintainers of projects using wtforms.

I would not be against making render_kw optional for now, with a deprecation warning for a definitive removal in the next release. This way we can share the burden.

Are you interested in opening a PR?

guidoiaquinti:
> About your point on old unmaintained projects breaking, and people having to spend time to update their code, well there is a tradeoff to be made: People ask for bugfixes and features in wtforms, that leads to refactoring, that sometimes leads to compatibility breakages. Keeping the old API puts the maintenance burden on the wtforms maintainers instead of the maintainers of projects using wtforms.
> 
> I would not be against making render_kw optional for now, with a deprecation warning for a definitive removal in the next release. This way we can share the burden.
> 
> Are you interested in opening a PR?

[...]

PR #816, opened 2023-10-26
Fix SelectField choice backwards compatibility

aminalaee:
Related to: https://github.com/wtforms/wtforms/issues/811

Fixes backwards compatiblity to SelectField choices to handle `render_kw` argument.


mapio:
This patch solves so many problems with other libraries using WTF (Flask ecosystem). So please accept it as soon as possible!

azmeuk:
I added a deprecation warning and released this under 3.1.1
Thank you for your feedbacks

PR #817, opened 2023-10-31
Fix SelectFieldBase.iter_choices docstring

alex-ball:
This fixes one of the out-of-date docstrings mentioned in #811 but not fixed in #812.


azmeuk:
Thank you.

PR #820, opened 2023-12-29
Fixes to SelectMultipleField

andycaine:
Make SelectMultipleField behave more like SelectField, as per the docs. Specifically:

- Do not override object data and default values when empty form data is provided.
- Allow coercible values to be provided as options.
- Do not require that coerced values are hashable.

### DIFF
--- a/src/wtforms/fields/choices.py
+++ b/src/wtforms/fields/choices.py
@@ -185,6 +185,9 @@ class SelectMultipleField(SelectField):
             self.data = None
 
     def process_formdata(self, valuelist):
+        if not valuelist:
+            return
+
         try:
             self.data = list(self.coerce(x) for x in valuelist)
         except ValueError as exc:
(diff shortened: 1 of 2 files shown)


davidism:
> Do not override object data and default values when empty form data is provided.

This is incorrect. Form data must provide values for every field, an empty value indicates no value at that point, not the object/default value.

davidism:
You've misunderstood how HTML forms and WTForms work together. `obj`/`data`/`default` populate the initial values in the HTML fields. When the HTML form is submitted, all values are sent, and represent the full data at that point.

andycaine:
> You've misunderstood how HTML forms and WTForms work together. `obj`/`data`/`default` populate the initial values in the HTML fields. When the HTML form is submitted, all inputs are sent, either with the values that may have been initially filled in or with user-provided values, and represent the full data at that point.

Yeah I get that.  But what about on the initial GET request?  Using the pattern above - form data is empty, so the fields should take their value from obj / data / default right?  SelectField and other fields do this, SelectMultipleField doesn't and I think it should.

davidism:
You don't pass the form data on get request, there is no form data. In fact, with Flask-WTF, this automatically happens with something along the lines of `if formdata is None and request.method == "POST": formdata = request.form`.

andycaine:
But the docs suggest that you do (e.g. the example above), and other fields accommodate this.  Why should SelectMutlipleField be different?

davidism:
I don't understand what you're trying to fix. You keep referring to the docs, and other behavior, but have shown neither. In fact, you've left those prompts blank in your pull request description. It's typically good to open an issue _before_ putting in work on a PR for this reason.

Issue #198, opened 2015-06-15
Handling JSON and nested form data


foxx:
I've attempted to use wtforms with JSON and nested `FormField` data, but it appears to be mostly unsupported. There are third party libraries such as [wtforms-json][1] which add support, but the code is quite dirty and of limited support. 

This is a use case which I'd imagine is quite popular, as wtforms is arguably the best form validation library out there for Python. 

Are there any plans to add support into the core for JSON? Could a core maintainer indicate whether such a feature request would be accepted? I'm happy to contribute time into a PR to make this happen. 

EDIT: It seems that in `2.0` there has already been some preparations towards this, such as the addition of `data` into `Form` constructor. 

[1]: https://github.com/kvesteri/wtforms-json

foxx:
Please ignore this issue, it seems that JSON is working perfectly fine with the `data` attribute of Form. If I get any further problems I'll re-open, apologies for the wasted time.

alanhamlett:
**DO NOT** pass user input to `data` or as a `kwarg` to a form because it is not validated! It's used as the default value for a field.

foxx:
Thanks for the reply, I was actually due to update this issue in the morning as I already ran into problems and noticed that validators/coercion was not working correctly. I attempted to use `obj` but this didn't work either. As it seems the original issue is still valid, I'll re-open this.

alanhamlett:
Yea, would be great to have this built into wtforms. Thanks for bringing this issue up!

aphillipo:
https://github.com/nicolaiarocci/cerberus 

Might be of interest if only dealing with APIs...

roganov:
This issue is very important, something should be done about supporting decoded JSON in WTForms. ATM event BooleanField doesn't work correctly (see https://github.com/wtforms/wtforms/blob/master/wtforms/fields/core.py#L692), not to mention FieldList and FormField.

@foxx May I ask what you ended up doing?

foxx:
@roganov I got rid of it and used [Colandar](http://docs.pylonsproject.org/projects/colander/en/latest/) instead. It takes a day of getting used to, but it's far more powerful.

crast:
Planned for WTForms 3.0 #154

foxx:
Also [schematics][1] is a good alternative, I prefer it more than Colandar.

[1]: https://schematics.readthedocs.org

azmeuk:
Is this feature still on the 3.0 roadmap?

[...]

PR #843, opened 2024-04-04
Make Optional check on field.data instead of field.raw_data

### LINKED ISSUE #842: Optional doesn't check field.data
If the form was filled with an object (like in REST scenario), `Optional` validator always raises `StopValidation` because it checks `field.raw_data` (and not the `field.data`) which will be always empty.

```python
from wtforms import Form
from wtforms.fields import StringField, FieldList
from wtforms.validators import IPAddress, Optional


class EditForm(Form):
    allow_list = FieldList(StringField(validators=[IPAddress(), Optional()]))


allow_list = ['555.555.555.555']


if __name__ == '__main__':
    form = EditForm()
    form.allow_list.process(None, allow_list)
    form.validate()
    print(form.errors)
```

### Actual Behavior
```python
> {}
```

### Expected Behavior
```python
> {'allow_list': [['Invalid IP address.']]}
```

### Possible fix

```python
class Optional:
    """
    Allows empty input and stops the validation chain from continuing.

    If input is empty, also removes prior errors (such as processing errors)
    from the field.

    :param strip_whitespace:
        If True (the default) also stop the validation chain on input which
        consists of only whitespace.

    Sets the `optional` attribute on widgets.
    """

    def __init__(self, strip_whitespace=True):
        if strip_whitespace:
            self.string_check = lambda s: s.strip()
        else:
            self.string_check = lambda s: s

        self.field_flags = {"optional": True}

        if (
            not field.data
            or isinstance(field.data, str)
            and not self.string_check(field.data)
        ):
            field.errors[:] = []
            raise StopValidation()
```
If `field.raw_data` will be replaced with `field.data`, validation works as expected. But I don't know if this fix will break something in other places.

### Environment

* Python version:  3.10.13
* wtforms version: 3.0.1

antonstakhouski:
When form is submitted via JSON and data is set using `field.process(None, data)`, `Optional` always stops the validation chain, because it checks only `field.raw_data` which will always be `None`.

Fix for issue #842 .

 `Optional` validator should check `field.data` to validate fields correctly.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -248,9 +248,9 @@ class Optional:
 
     def __call__(self, form, field):
         if (
-            not field.raw_data
-            or isinstance(field.raw_data[0], str)
-            and not self.string_check(field.raw_data[0])
+            not field.data
+            or isinstance(field.data, str)
+            and not self.string_check(field.data)
         ):
             field.errors[:] = []
             raise StopValidation()
--- a/tests/validators/test_optional.py
+++ b/tests/validators/test_optional.py
@@ -37,3 +37,19 @@ def test_input_optional_raises(data_v, raw_data_v, dummy_form, dummy_field):
         validator(dummy_form, dummy_field)
 
     assert len(dummy_field.errors) == 0
+
+
+def test_input_optional_raises_with_empty_formdata(dummy_form, dummy_field):
+    """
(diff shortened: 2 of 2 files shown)


azmeuk:
This changes the semantics of Optional. In a future major version with breaking changes we may rethink this part, but not for now.

Somewhat related to #198

PR #844, opened 2024-05-07
InputRequired send error only when data is None

antoniosobrinho:
InputRequired return error when data is False or 0.

If my data is a Int or a Bool and I need them to be required, InputRequired  return error if I send False or 0, but I send the data.

This fix check only if data is None, so, if the data exists

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -316,7 +316,7 @@ class InputRequired:
         self.field_flags = {"required": True}
 
     def __call__(self, form, field):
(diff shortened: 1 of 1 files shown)


Daverball:
`raw_data` should always be a list of strings (or a file uploads, the representation of which depends on if you use Webob, Werkzeug or something else to parse your POST/GET formdata), if you pass in something else then you are not using the library correctly.

`DataRequired` actually has the issue you're trying to fix here, but `InputRequired` does not.

PR #868, opened 2024-11-08
Fix: boolean field not using force_values when the form initializes it.

jeffdico:
The booleanField is not using the false_values to check the formdata value  passed to it. it just sets what ever vallue passed as bool.

This patch makes a call to process_formdata method so as to  enforce using false_values set when creating the field in the form class.

### DIFF
--- a/src/wtforms/fields/simple.py
+++ b/src/wtforms/fields/simple.py
@@ -40,7 +40,7 @@ class BooleanField(Field):
 
     def process_data(self, value):
         # this will initialize the bool fields with false_values before setting it
-        self.process_formdata([value]) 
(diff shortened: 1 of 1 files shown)


azmeuk:
>  The booleanField is not using the false_values to check the formdata value passed to it. it just sets what ever vallue passed as bool.

But it does:

https://github.com/pallets-eco/wtforms/blob/bdc8121699ce67cbb1e631b64551a75f1eb8e125/src/wtforms/fields/simple.py#L45-L49

Closing for now as your diagnostic is not clear. Please describe your use case and provide a minimum reproducible example that can be copy/pasted in a REPL as the issue template suggests.

PR #870, opened 2024-11-21
Select widget: Add support for mixture of grouped and non-grouped choices

Lukas0907:
Currently it's only possible to have all select `option`s inside `optgroup`s or not use `optgroup`s at all. This PR adds support for allowing non-grouped options (which appear next to groups) as well.


azmeuk:
Superseded by #889

PR #871, opened 2024-11-22
Add "search", "match" and "fullmatch" modes to Regexp validator

### LINKED ISSUE #867: It would be great if `validators.regexp` could use `re.fullmatch`
The `regexp` validator uses `re.match`:

https://github.com/pallets-eco/wtforms/blob/bbfb78c487a2d95dd012a6332d719a3d440498f9/src/wtforms/validators.py#L352

This means that it's not trivial to ensure that the _entire string_ matches a regex, especially when that regex is complex.

I propose something like this:

```python
class Regexp:
    """
    Validates the field against a user provided regexp.

    :param regex:
        The regular expression string to use. Can also be a compiled regular
        expression pattern.
    :param flags:
        The regexp flags to use, for example re.IGNORECASE. Ignored if
        `regex` is not a string.
    :param message:
        Error message to raise in case of a validation error.
    :param re_match_type
        "match" (default), "fullmatch", or "search"
    """

    def __init__(self, regex, flags=0, message=None, re_match_type="match":
        if isinstance(regex, str):
            regex = re.compile(regex, flags)
        self.regex = regex
        self.message = message
        self.re_match_type = re_match_type

    def __call__(self, form, field, message=None):
        if self.re_match_type == "match":
            match = self.regex.match(field.data or "")
        if self.re_match_type == "fullmatch":
            match = self.regex.fullmatch(field.data or "")
        elif self.re_match_type == "search":
            match = self.regex.search(field.data or "")
        else:
            raise ValueError("... but how???")
            
        if match:
            return match

        if message is None:
            if self.message is None:
                message = field.gettext("Invalid input.")
            else:
                message = self.message

        raise ValidationError(message)
```

Obviously I could make it more "intelligent" with `getattr` or something, and a real implementation would probably use an enum and do the validation elsewhere, but for the purposes of getting the idea across, the above is sufficient.

Thoughts? Would you mind if I implemented this?

### Environment

* Python version: 3.12
* wtforms version: 3.1.2

artempronevskiy:
Description:

Currently `Regexp` validator supports only `re.match` mode, but sometimes we need to use `re.fullmatch` or `re.search` mode.

Relevant issues or pull requests:

https://github.com/pallets-eco/wtforms/issues/867

What this PR does to fix the issue:

- Added support of `re.fullmatch` and `re.search` modes along with `re.match` with full backward compatibility
- Introduced new `ValidatorSetupError` exception to be raised if an user is trying to set up validator in a wrong way
- Added additional tests for `Regexp` validator to cover the changes
- Updated `docs/fields.rst` file to show added `mode` kwarg to `Regexp.__init__()`

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -233,7 +233,7 @@ refer to a single input from the form.
     Example usage::
 
         class UploadForm(Form):
-            image        = FileField('Image File', [validators.regexp('^[^/\\]\.jpg$')])
+            image        = FileField('Image File', [validators.regexp('^[^/\\]\.jpg$', mode='fullmatch')])
             description  = TextAreaField('Image Description')
 
             def validate_image(form, field):
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -32,12 +32,22 @@ __all__ = (
     "mac_address",
     "UUID",
     "ValidationError",
+    "ValidatorSetupError",
     "StopValidation",
     "readonly",
     "ReadOnly",
     "disabled",
     "Disabled",
 )
+
+
+class ValidatorSetupError(ValueError):
+    """
+    Raised when a validator is configured improperly.
+    """
+
+    def __init__(self, message="", *args, **kwargs):
+        ValueError.__init__(self, message, *args, **kwargs)
 
 
 class ValidationError(ValueError):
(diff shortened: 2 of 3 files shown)


azmeuk:
Superseded by https://github.com/pallets-eco/wtforms/pull/918

PR #873, opened 2024-12-19
Fix Length Validator passing `None` when min=0

### LINKED ISSUE #872: Length Validator Broken when min=0
PR incoming.

### Actual Behavior

WTForm's `Length` validator incorrectly approves a missing string (aka `field.data` is `None`) when `min=0`, whereas it should fail validation.

I interpret `min=-1` to mean 'no minimum length, permit `None` and the empty string `''`'

I interpret `min=0` to mean 'minimum length 0, do not permit `None`, aka a missing field, but do permit the empty string `''`'

Serving the following app with `$ gunicorn min-len-fail:app` on the default localhost:8000

```python
from flask              import Flask, request, jsonify
from wtforms            import Form, StringField
from wtforms.validators import Length

app = Flask(__name__)

class TestForm(Form):
    title = StringField("Title", validators = [Length(min=0)])

@app.route("/", methods = ["POST"])
def index():
    form = TestForm(request.form)
    if form.validate():
        return jsonify({"msg": "Success!", "title": form.title.data})
    else:
        return jsonify(form.errors), 400
```

Test POSTing `title` both as missing, and as the empty string, with curl

```
$ curl -sSL -d 'titl=' localhost:8000
{"msg":"Success!","title":null}
$ curl -sSL -d 'title=' localhost:8000
{"msg":"Success!","title":""}
```

The Length validator reports success for both cases, whereas it should fail the first.

### Expected Behavior

With the PR incoming

```
$ curl -sSL -d 'titl=' localhost:8000
{"title":["Field must be at least 0 characters long."]}
$ curl -sSL -d 'title=' localhost:8000
{"msg":"Success!","title":""}
```

### Environment

* Python version: 3.12.7, Arch Linux
* wtforms version: main branch (3.2.1)

### Cause

```py
class Length:
    ...
        length = field.data and len(field.data) or 0
        ...
```

The default length of `0` is incorrect; it should be `-1`. And no we can't just change the `0` to a `-1` in that awful `x and y or z` mess otherwise the empty string would incorrectly set `length` to `-1`.

I've added three tests for the `Length` validator to stress-test the difference between the `None` missing string and the `''` empty string.

jb2170:
Closes #872 

I don't think any fields other than `StringField` use the Length validator, so there's no need for an extra `elif not field.data: length = 0` line to apply to falsey values: Truthy data likewise would need to implement the `len` operator in the pre-existing implementation which only string data does. So everything is good.

I've included three stress-test / edge-case scenarios that make the distinction between `None` and `''` clear. The pre-existing implementation fails these tests, and with this PR they pass.

### DIFF
--- a/src/wtforms/validators.py
+++ b/src/wtforms/validators.py
@@ -134,7 +134,11 @@ class Length:
             self.field_flags["maxlength"] = self.max
 
     def __call__(self, form, field):
-        length = field.data and len(field.data) or 0
+        if field.data is None:
+            length = -1
+        else:
+            length = len(field.data)
+
         if length >= self.min and (self.max == -1 or length <= self.max):
             return
 
--- a/tests/validators/test_length.py
+++ b/tests/validators/test_length.py
@@ -58,6 +58,7 @@ def test_length_messages(dummy_form, dummy_field, validator, message):
 
     assert message in str(e.value)
 
+
 @pytest.mark.parametrize("min_v, max_v", [(-1, 5)])
 def test_null_str_pass(min_v, max_v, dummy_form, dummy_field):
     """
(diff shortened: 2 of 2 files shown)


jb2170:
Closed due to breaking changes: forking off instead. See [issue comment](https://github.com/pallets-eco/wtforms/issues/872#issuecomment-2778368678).

PR #880, opened 2025-08-21
fix(validators): Disabled validation with provided formdata

subnix:
The `Field.raw_data` attribute has three states:

1. `None`: The form data hasn't been passed.
2. A list with a single value: The form data has been passed, and the field is present.
3. An empty list: The form data has been passed, but the field is missing.

https://github.com/pallets-eco/wtforms/blob/7477295a479b77aff322796c4a0c7b7753ce3333/src/wtforms/fields/core.py#L21

https://github.com/pallets-eco/wtforms/blob/7477295a479b77aff322796c4a0c7b7753ce3333/src/wtforms/fields/core.py#L313-L317

Currently, the validator correctly supports the first two states (passes the validation and fails, respectively). However, it should also support the third state, which should pass the validation.

Relates to #792  #797

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: wtforms
 
+Unreleased
+-------------
+
+Unreleased
+
+- Fix :class:`~validators.Disabled` validation with provided formdata. :pr:`880`
+
 Version 3.2.1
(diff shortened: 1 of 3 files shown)


azmeuk:

PR #884, opened 2026-01-15
Update FAQ to reflect 3.10+ support

kurtmckee:
While reviewing the docs I noticed that the FAQ page says Python 3.9+ is supported. However, a recent commit bumped the minimum Python to 3.10.

This PR updates the FAQ to reflect the newly-required minimum version. :+1:


azmeuk:

Issue #922, opened 2026-05-10
Release 3.3.0 incoming


azmeuk:
Hi everyone,

I worked on all the opened issues and PR that don't required overhauling the library. I released version 3.3.0b1 and will release version 3.3.0 final in the coming days or weeks.
Please give feedback here if you notice things that deserve changes before 3.3.0 final. We can still change the API. I will try to reach a few maintainers of projects depending on wtforms before cutting the release.

[The full changelog is here](https://wtforms.readthedocs.io/en/main/changes/#version-3-3-0b1). Most changes are non breaking but those points may require attention:
- the (re)introduction of the `SelectChoice` dataclass might break fields inheriting SelectField and expecting `iter_choices` to return a list of tuples, now this is a list of `SelectChoice`.
- `SelectField` no longer has `has_groups` and `iter_groups`
- `URL` no longer inherits `Regexp`, and use urllib internally to parse urls. It accepts only http and https by default, other schemes must be whitelisted with schemes=
- `FloatField` default widget is `NumberInput` and no longer `TextInput`

After this, and I won't promise exactly when, I will work on the remaining tickets for wtforms v4. This version will have breaking changes as I want to rework the form processing part so it is more consistent, the field binding part etc. It should bring interesting features too, such as type annotations #618, multiple fields #799, maybe async #839.

There will maybe be an intermediate 3.4 release containing compatibility helpers for v4 and deprecation warnings.

/cc @PanderMusubi @Daverball @ThiefMaster @davidism @TurnrDev

Daverball:
I'm a little bit concerned about the breaking change `SelectChoice` introduces, especially because the current design feels a little sloppy with the internal `_selected` field that's used to smuggle that value through `iter_choices`.

I think I would like it better if greater care was taken to keep things somewhat more backwards compatible for custom `SelectField` subclasses, especially since it's quite common for `choices` to be populated after the field is constructed, if the choices are dynamic. You took care to keep the static path compatible, but the dynamic path will break still.

Here's a few suggestions for an alternate design:
1. Have separate types for the choices that are stored on the field, and the ones that are returned by `iter_choices`, i.e. the `selected` field only exists on the type returned by `iter_choices`.
2. Make those types `NamedTuple` objects, so they can still be used in the same way as the old raw tuples, while retaining the ergonomic benefits of giving names to each component
[...]

Issue #886, opened 2026-03-10
Support dict for SelectForm.choices


marksteward:
A list of tuples made sense before dicts were ordered by default, but we're now already using dicts for optgroup support. It'd be nice to be able to do:

```
class PastebinEntry(Form):
    language = SelectField(
        'Programming Language',
        choices={
            'cpp': 'C++',
            'py': 'Python',
            'text': 'Plain Text',
        },
    )

form = PastebinEntry()
del form.choices['cpp']  # instead of having to iterate over .choices and filter out 'cpp'
```

This could still coexist with an optgroup version:

```
class PastebinEntry(Form):
    language = SelectField(
        'Programming Language',
        choices={
            'Actual languages': {'cpp': 'C++', 'py': 'Python'},
            'Generic': {'text': 'Plain Text'},
        },
    )
```

or the current allowed format:

```
class PastebinEntry(Form):
    language = SelectField(
        'Programming Language',
        choices={
            'Actual languages': [('cpp', 'C++'), ('py', 'Python')],
            'Generic': [('text', 'Plain Text')],
        },
    )
```

Happy to provide a PR if this sounds sensible!

marksteward:
I'll take that as PR not welcome, then.

azmeuk:
Hi,
Thank you for your suggestion and sorry for not having answered it.
dicts indeed appears more elegant than lists of tuples we use to have, #889 (re)introduced a specialized `Choice` dataclass which goes further than dict can, for instance by supporting render_kw.
Giving it a second thought, that would probably not be incompatible with choice dicts, but I wonder if we want to support several APIs for choices.

azmeuk:
For the record, I added support for dict choices in #923

Issue #618, opened 2020-06-10
Add type annotations


benbariteau:
It would be nice to be able to use typecheckers like mypy with wtforms.

diegor8:
Any progress on this issue?

miketheman:
Is this change desired? Would love to see a comment from a maintainer if they'd entertain accepting a PR adding type hints before putting time in on trying to make it happen.

davidism:
I am entirely neutral on adding them, but I probably wouldn't accept a PR from someone else. From my experience adding typing to Flask, etc., I have particular standards I want to meet for the initial annotations, there's a lot of complexity in the implementation of WTForms bound fields, and it's very hard to review a PR that changes the majority of lines.

Daverball:
There's now a third-party stubs package: https://pypi.org/project/types-WTForms/

There are some caveats, since the API has some rough edges which makes it hard to properly represent in the current type system, feel free to look through some of my comments in the original PR: https://github.com/python/typeshed/pull/10557

The main issue is that `Field.__new__` usually returns `UnboundField`, but mypy currently only allows returning a subclass and if you violate that rule it will ignore the return type and assume it's `Field` anyways. There is an open PR in mypy to change that behavior, but it's still not entirely clear how type checkers will behave with both `__new__` and `__init__` present, so right now for convenience's sake the API just accepts `Field` for some arguments that should only accept `UnboundField` (specifically `FieldList` is affected)

jace:
I don't have `types-WTForms` installed, but Pylance appears to be using it anyway and is redlining my code. In particular:

1. `LazyString` from Flask-Babel is not compatible with `str`. The workaround is to re-cast the type on import:

    ```python
    from typing import Callable
    from flask_babel import lazy_gettext
    __: Callable[..., str] = lazy_gettext
    ```

2. I have a custom field that types `validators` as `list[Callable[[Form, Field], None]]` with `Form` and `Field` imported from WTForms. This is now erroring on the `URL()` validator:

    ```
    Argument of type "list[URL]" cannot be assigned to parameter "validators" of type "ValidatorList" in function "__init__"
      Type "URL" cannot be assigned to type "ValidatorCallable"
        Parameter 2: type "Field" cannot be assigned to type "StringField"
          "Field" is incompatible with "StringField"
    ```

    That last line is suggesting the type stub's `StringField` has a different base from WTForms's `Field`.

[...]

PR #923, opened 2026-05-13
SelectField backward compatibility and minor QoL improvements

azmeuk:
I addressed most of the points raised in #922

- Add a `post_process` step which happens right after `process`
- Add optional `form` and `field` args to `SelectField` and `DataList` choices callbacks. Callbacks are resolved during the `post_process` step, so it can access the other fields `.data`
- `Choice` inherit from `NamedTuple` instead of `Dataclass`. An additional private `_Choice` class is needed to pass default values to the tuple members since NamedTuple forbids overriding __new__
- Restore `has_groups` and `iter_groups`
- Backward compatibility for `iter_choices` returning tuples, for `render_options` taking `value, label, selected` instead of `choice`.
- Support choices as dict as suggested in #886

The compatibility layer makes the code bigger and harder to read, but hopefully things goes better when we delete the deprecations in 3.4/4.0.

I have tested those modification in downstream projects (wtf-peewee, wtforms-alchemy, starlette-wtf, flask-appbuilder) and it solves all the issues related to SelectField.

/cc @daverball

### DIFF
--- a/src/wtforms/fields/choices.py
+++ b/src/wtforms/fields/choices.py
@@ -1,7 +1,11 @@
 import warnings
 from dataclasses import dataclass
 from dataclasses import field
+from dataclasses import replace
 from enum import Enum
+from itertools import groupby
+from operator import attrgetter
+from typing import NamedTuple
 
 from wtforms import widgets
 from wtforms._compat import get_signature
@@ -29,24 +33,62 @@ def _enum_coerce(enum_cls):
     return coerce
 
 
-@dataclass
-class Choice:
+class Choice(NamedTuple):
     """
-    A dataclass that represents an available choice for choice fields.
+    A rendered option yielded by
+    :meth:`SelectFieldBase.iter_choices` and
+    :meth:`SelectFieldBase.iter_groups`.
+
+    ``selected`` is computed against the field's current data. To
+    declare options on a :class:`SelectField`, use
+    :class:`SelectChoice` instead.
 
     :param value:
         The value that will be sent in the request.
     :param label:
         The label of the option.
+    :param selected:
+        Whether the option is currently selected. Set by ``iter_choices``;
+        you rarely set this yourself.
     :param render_kw:
         A dict containing HTML attributes that will be rendered
         with the option.
     """
 
     value: str
-    label: str | None = None
-    render_kw: dict | None = None
-    _selected: bool = field(default=False, kw_only=True)
+    label: str
+    selected: bool
+    render_kw: dict
+
+
+@dataclass
+class SelectChoice:
+    """
+    An option declared via :class:`SelectField` and
+    :class:`SelectMultipleField`'s ``choices=`` parameter.
+
+    :param value:
+        The value that will be sent in the request.
+    :param label:
+        The label of the option. Defaults to ``value`` when omitted.
+    :param render_kw:
+        A dict containing HTML attributes that will be rendered
+        with the option. Defaults to an empty dict when omitted.
+    :param optgroup:
+        The ``<optgroup>`` HTML tag in which the option will be rendered.
+    """
+
+    value: str
+    label: str = None  # type: ignore[assignment]
+    render_kw: dict = field(default_factory=dict)
+    optgroup: str | None = None
+
+    def __post_init__(self):
+        if self.label is None:
+            self.label = self.value
+
+    def __iter__(self):
+        return iter((self.value, self.label, self.render_kw, self.optgroup))
 
     @classmethod
     def from_enum(cls, enum_cls, *, label=None):
@@ -61,45 +103,41 @@ class Choice:
             label = str if "__str__" in enum_cls.__dict__ else lambda m: m.name
         return [cls(value=m.name, label=label(m)) for m in enum_cls]
 
-
-@dataclass
-class SelectChoice(Choice):
-    """
-    A :class:`Choice` augmented with an ``<optgroup>`` hint for
-    :class:`SelectField`.
-
-    :param optgroup:
-        The ``<optgroup>`` HTML tag in which the option will be rendered.
-    """
-
-    optgroup: str | None = None
-
     @classmethod
     def from_input(cls, input, optgroup=None):
+        """Coerce a value passed by the user via ``choices=...`` into a
+        :class:`SelectChoice`.
+        """
         if isinstance(input, SelectChoice):
             if optgroup:
-                input.optgroup = optgroup
+                return replace(input, optgroup=optgroup)
             return input
 
         if isinstance(input, Choice):
+            warnings.warn(
+                "Passing Choice to a SelectField is deprecated; Choice is the "
+                "output type returned by iter_choices(). Use SelectChoice "
+                "instead. Support for Choice as input will be removed in "
+                "WTForms 4.0.",
+                DeprecationWarning,
+                stacklevel=4,
+            )
             return cls(
                 value=input.value,
                 label=input.label,
                 render_kw=input.render_kw,
                 optgroup=optgroup,
-                _selected=input._selected,
             )
 
         if isinstance(input, str):
             return cls(value=input, optgroup=optgroup)
 
         if isinstance(input, tuple):
-            warnings.warn(
-                "Passing SelectField choices as tuples is deprecated and will be "
-                "removed in wtforms 3.4. Please use Choice or SelectChoice instead.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
+            if len(input) not in (2, 3):
+                raise ValueError(
+                    f"SelectField choice tuple must have 2 or 3 elements, "
+                    f"got {len(input)}"
+                )
             return cls(*input, optgroup=optgroup)
 
 
@@ -120,10 +158,18 @@ class SelectFieldBase(Field):
             self.option_widget = option_widget
 
     def iter_choices(self):
+        """Provide data for choice widget rendering.
+
+        Should yield :class:`Choice` instances.
         """
-        Provides data for choice widget rendering. Must return a sequence or
-        iterable of SelectChoice.
-        """
+        raise NotImplementedError()
+
+    def has_groups(self):
+        """Whether the field's choices include any ``optgroup`` hint."""
+        return False
+
+    def iter_groups(self):
+        """Yield ``(group_label, [Choice, ...])`` pairs for grouped rendering."""
         raise NotImplementedError()
 
     def __iter__(self):
@@ -142,11 +188,12 @@ class SelectFieldBase(Field):
                 **opts,
             )
             opt.choice = choice
-            opt.checked = choice._selected
+            opt.checked = choice.selected
             opt.process(None, choice.value)
             yield opt
 
-    def choices_from_input(self, choices):
+    def _choices_from_input(self, choices):
+        """Parse the user-supplied ``choices`` into a list of :class:`SelectChoice`."""
         if callable(choices):
             choices = self._invoke_choices_callback(choices)
 
@@ -154,14 +201,6 @@ class SelectFieldBase(Field):
             return None
 
         if isinstance(choices, dict):
-            warnings.warn(
-                "Passing SelectField choices in a dict deprecated and will be removed "
-                "in wtforms 3.4. Please pass a list of SelectChoice objects with a "
-                "custom optgroup attribute instead.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-
             return [
                 SelectChoice.from_input(input, optgroup)
                 for optgroup, inputs in choices.items()
@@ -209,7 +248,7 @@ class SelectField(SelectFieldBase):
             self.choices = None
         else:
             self._choices_callable = None
-            self.choices = self.choices_from_input(choices)
+            self.choices = self._choices_from_input(choices)
         self.validate_choice = validate_choice
         self.invalid_value_message = invalid_value_message or self.gettext(
             "Invalid Choice: could not coerce."
@@ -219,15 +258,41 @@ class SelectField(SelectFieldBase):
         )
 
     def iter_choices(self):
-        choices = self.choices_from_input(self.choices) or []
-        for choice in choices:
-            choice._selected = self.coerce(choice.value) == self.data
-        return choices
+        choices = self._choices_from_input(self.choices) or []
+        return [
+            Choice(
+                value=c.value,
+                label=c.label,
+                selected=self.coerce(c.value) == self.data,
+                render_kw=c.render_kw,
+            )
+            for c in choices
+        ]
+
+    def has_groups(self):
+        choices = self._choices_from_input(self.choices) or []
+        return any(c.optgroup is not None for c in choices)
+
+    def iter_groups(self):
+        choices = self._choices_from_input(self.choices) or []
+        for optgroup, group in groupby(choices, key=attrgetter("optgroup")):
+            yield (
+                optgroup,
+                [
+                    Choice(
+                        value=c.value,
+                        label=c.label,
+                        selected=self.coerce(c.value) == self.data,
+                        render_kw=c.render_kw,
+                    )
+                    for c in group
+                ],
+            )
 
     def post_process(self):
         super().post_process()
         if self._choices_callable is not None:
-            self.choices = self.choices_from_input(self._choices_callable)
+            self.choices = self._invoke_choices_callback(self._choices_callable)
 
     def process_data(self, value):
         try:
@@ -255,7 +320,7 @@ class SelectField(SelectFieldBase):
         if self.choices is None:
             raise TypeError(self.gettext("Choices cannot be None."))
 
-        if not any(choice._selected for choice in self.iter_choices()):
+        if not any(choice.selected for choice in self.iter_choices()):
             raise ValidationError(self.invalid_choice_message)
 
 
@@ -298,11 +363,34 @@ class SelectMultipleField(SelectField):
         self.invalid_choice_message = invalid_choice_message
 
     def iter_choices(self):
-        choices = self.choices_from_input(self.choices) or []
-        if self.data:
-            for choice in choices:
-                choice._selected = self.coerce(choice.value) in self.data
-        return choices
+        choices = self._choices_from_input(self.choices) or []
+        data = self.data or ()
+        return [
+            Choice(
+                value=c.value,
+                label=c.label,
+                selected=self.coerce(c.value) in data,
+                render_kw=c.render_kw,
+            )
+            for c in choices
+        ]
+
+    def iter_groups(self):
+        choices = self._choices_from_input(self.choices) or []
+        data = self.data or ()
+        for optgroup, group in groupby(choices, key=attrgetter("optgroup")):
+            yield (
+                optgroup,
+                [
+                    Choice(
+                        value=c.value,
+                        label=c.label,
+                        selected=self.coerce(c.value) in data,
+                        render_kw=c.render_kw,
+                    )
+                    for c in group
+                ],
+            )
 
     def process_data(self, value):
         try:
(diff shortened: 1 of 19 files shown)


ElLorans:
Couldn't double check everything, but if we were extending a Field now we have to change from `iter_choices` to `_iter_choices_normalized`?

src/wtforms/fields/choices.py
```diff
@@ -169,6 +262,55 @@ def choices_from_input(self, choices):
...
+                    "will be removed in wtforms 3.4. Please use Choice or "
+                    "SelectChoice instead.",
+                    DeprecationWarning,
+                    stacklevel=3,
+                )
+                return
+
+    def _invoke_choices_callback(self, cb):
+        try:
+            sig = inspect.signature(cb)
```
Daverball: For better compatibility with Python 3.14 I recommend extracting this into a helper function in a `compat` module:

```
import inspect
import sys

if sys.version_info < (3, 14):
    def get_signature(fn):
        return inspect.signature(fn)
else:
    from annotationlib import Format

    def get_signature(fn):
         return inspect.signature(fn, annotation_format=Format.FORWARDREF)
```

Otherwise you risk breaking working code for people that want to migrate away from `from __future__ import annotations`.

Daverball (commented):
Thanks for working on this, this looks pretty good and alleviates my concerns. I will try to take another deep dive later, to see if there is anything else you or I missed.

I did spot one small issue with the new callback, but that's an easy fix, it just adds a little bit more code.

azmeuk:
> Couldn't double check everything, but if we were extending a Field now we have to change from iter_choices to _iter_choices_normalized?

No, you keep overriding iter_choices. _iter_choices_normalized is just a private helper that normalizes the iter_choices output to keep backward compatibility (and raise deprecation warnings).

src/wtforms/fields/choices.py
azmeuk: Good catch.

src/wtforms/fields/choices.py
```diff
@@ -169,6 +269,55 @@ def choices_from_input(self, choices):
...
+    def _warn_legacy_choices(choices):
+        """Emit a one-shot ``DeprecationWarning`` for legacy ``choices`` shapes.
+
+        Legacy shapes are raw tuples or ``dict``.
+        """
+        if isinstance(choices, dict):
+            if SelectFieldBase._is_shorthand_dict(choices):
+                return
+            warnings.warn(
+                "Passing SelectField choices in a dict deprecated and will be "
```
Daverball:
```suggestion
                "Passing SelectField choices in a dict is deprecated and will be "
```

src/wtforms/fields/choices.py
```diff
@@ -202,10 +362,33 @@ def __init__(
...
+        ungrouped = []
+        for c in self._iter_choices_normalized():
+            item = Choice(c.value, c.label, c.selected, c.render_kw)
+            if c.optgroup is None:
+                ungrouped.append(item)
+            else:
+                groups.setdefault(c.optgroup, []).append(item)
+        yield from groups.items()
+        if ungrouped:
+            yield None, ungrouped
```
Daverball: I think the behavior of this with SelectChoice will seem a little surprising to people used to how itertools.grouped works.
Let's say you have these four choice, group pairs:
```
A, foo
B, baz
C, None
D, foo
```

My intuition tells me I will end up with three `optgroup`, each consisting of a single `option` and one ungrouped `option` in between. Two of those `optgroup` will have the same name foo.
Your current implementation will reorder choices. I would personally suggest to just use `itertools.grouped` here, while what I described should be rare and somewhat esoteric, there's no reason to make things more complicated, than they need to be.

src/wtforms/form.py
```diff
@@ -246,6 +267,7 @@ def __init__(
         prefix="",
         data=None,
         meta=None,
+        _parent_form=None,
```
Daverball: I'm a little concerned about custom `Form` subclasses getting broken by this extra private parameter. `Form` has few enough parameters that I can see people duplicating the signature in subclasses, rather than using `*args, **kwargs`, especially when using type annotations. Could we get away with `FormField` just setting this private attribute after creating the form instance, instead of making it part of the constructor contract?

src/wtforms/fields/choices.py
```diff
@@ -202,10 +362,33 @@ def __init__(
...
     def iter_choices(self):
-        choices = self.choices_from_input(self.choices) or []
-        for choice in choices:
-            choice._selected = self.coerce(choice.value) == self.data
-        return choices
+        choices = self._choices_from_input(self.choices) or []
+        return [
+            choice._replace(selected=self.coerce(choice.value) == self.data)
+            for choice in choices
+        ]
```
Daverball, line 370: I didn't catch this in my first review, but with extra `optgroup` portion in the `NamedTuple` some users of `iter_choices` and `iter_groups` might still crash, if they don't expect the fifth component in the `tuple` and aren't set up to deal with that, so instead of using `_replace` here, I think you should construct a `Choice` object


```suggestion
        choices = self._choices_from_input(self.choices) or []
        return [
            Choice(
                value=choice.value,
                label=choice.label,
                selected=self.coerce(choice.value) == self.data,
                render_kw=choice.render_kw
            )
            for choice in choices
        ]
```

It would also make sense to me to remove `selected` from `SelectChoice`, I'm still not super happy that `selected` is exposed outside of `iter_choices`, it doesn't really make sense in the context of user supplying choices to a `SelectField`, since its value will be ignored and replaced, when rendering the choices. And I'm not sure whether I would allow people passing `Choice` objects into `SelectField`, it might be less confusing if they're documented as internal objects returned by `iter_choices` and `iter_groups`.

src/wtforms/fields/choices.py
Daverball: I think you also forgot to calculate `selected` here, so I guess this should actually be this:


```suggestion
        for name, group in grouped(
            self._iter_choices_normalized(),
            key=attrgetter('optgroup')
        ):
            yield name, [
                Choice(
                    value=choice.value,
                    label=choice.label,
                    selected=self.coerce(choice.value) == self.data,
                    render_kw=choice.render_kw
                )
                for choice in group
            ]
```

Daverball (commented):
Upon closer review I found a few additional subtle issues, but other than that I think we're good.

src/wtforms/widgets/core.py
```diff
@@ -438,11 +443,49 @@ def render_option(cls, choice, **kwargs):
...
 
+    @classmethod
+    def _dispatch_render_option(cls):
+        """Return a callable ``(choice) -> str`` that invokes :meth:`render_option`.
+
+        Picks the right signature for the override.
+        """
+        ro = cls.render_option
+        try:
+            sig = inspect.signature(ro)
```
Daverball: I missed this other use of `inspect.signature`. This has the same problem as the other use.

src/wtforms/widgets/core.py
```diff
@@ -420,13 +423,15 @@ def __call__(self, field, **kwargs):
                 kwargs[k] = getattr(flags, k)
         select_params = html_params(name=field.name, **kwargs)
         html = [f"<select {select_params}>"]
-        choice_groups = self.sort_by_optgroup(field.iter_choices())
+        iter_choices = getattr(field, "_iter_choices_normalized", field.iter_choices)
```
Daverball: This seems a little fragile, I would recommend still using the old interface of `iter_choices`, `has_groups` and `iter_groups` here, in case a subclass overrides those methods, without overwriting `_iter_choices_normalized`.

You could potentially check if `iter_choices` and `iter_groups` are still the original methods and haven't been overwritten and dispatch to the new internal method as a speed optimization in the most common case, when you know nothing has been changed. But I'm not sure it's worth the complexity.

Also `sort_by_optgroup` has the same problems I outline in `iter_groups`. So I still think it makes more sense to not duplicate all this logic into the widget and instead just rely on the public interface. You can still change the preferred signature of `render_option` and emit a deprecation warning for the old signature.

Daverball (commented):
I stand corrected, found a couple more things in the widget code.

src/wtforms/form.py
azmeuk: Form.__init__ calls Form.process, which needs _parent_form to know if `post_process` should be triggered or not, and this to avoid that subforms (in FieldForm) run `post_process` multiple times.

I could avoid the additional __init__ parameter by adding this in the metaclass.

Daverball: You're right, I missed that part. I think adding it to the metaclass would be better, since it's more backwards compatible.

src/wtforms/widgets/core.py
```diff
@@ -438,17 +446,48 @@ def render_option(cls, choice, **kwargs):
...
+            "render_option(cls, choice, **kwargs) instead — choice is a "
+            "SelectChoice. The legacy signature will be removed in WTForms "
+            "4.0.",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+        accepts_kwargs = any(p.kind == p.VAR_KEYWORD for p in sig.parameters.values())
+
+        def adapter(choice):
+            args = (choice.value, choice.label or choice.value, choice.selected)
```
Daverball, line 485: ~Since `choice` is a `NamedTuple` we can use it directly as the `args` `tuple`, this also means subclasses that still return a raw `tuple` will not yet be broken by this adapter.~ It would probably also be a good idea to emit a `DeprecationWarning` if `choice` is a raw `tuple` and coerce to `Choice`, so you can safely use the named attributes in rendering.

Daverball: Actually, nevermind, I missed the `render_kw` part of this. So you can't directly use it. The other part of the comment still stands though.

Daverball: You may need to put the check somewhere else though, so it also happens for the default renderer with the new signature, or duplicate the check into the new renderer, although that's a little more fragile for custom renderers.

src/wtforms/fields/choices.py
```diff
@@ -119,48 +195,67 @@ def __init__(self, label=None, validators=None, option_widget=None, **kwargs):
             self.option_widget = option_widget
 
     def iter_choices(self):
-        """
-        Provides data for choice widget rendering. Must return a sequence or
-        iterable of SelectChoice.
+        """Provide data for choice widget rendering.
+
+        Should yield :class:`SelectChoice` instances.
```
Daverball:
```suggestion
        Should yield :class:`Choice` instances.
```
Stale documentation

src/wtforms/widgets/core.py
azmeuk: > It would probably also be a good idea to emit a DeprecationWarning if choice is a raw tuple and coerce to Choice, so you can safely use the named attributes in rendering.

The check is already handled upstream of the widget in `_normalize_choice` and emits a warning for raw tuples.

Daverball: Isn't it possible to bypass that check if they override `iter_choices` or `iter_groups`? Before the change it wasn't possible because you were using the internal helper method to iterate the choices, but now that you're directly using the old interface, you need to be careful on the consumer-side about the shape of the choices as well.

src/wtforms/fields/choices.py
azmeuk: Now there is SelectChoice/DataListChoice for inputs, and Choice as inner model.

azmeuk:
I think I adressed most of your feedback. I went back to dataclasses with __iter__ for the tuple compatibility, because this is hard to handle default values and inheritance with them. I tested on several dowstream projects and the unit test suites pass.

Daverball:
> I think I adressed most of your feedback. I went back to dataclasses with **iter** for the tuple compatibility, because this is hard to handle default values and inheritance with them. I tested on several dowstream projects and the unit test suites pass.

Thanks so much, I'll take another closer look later, but there's one thing I can already say now.

Using dataclasses for `SelectChoice`/`DataListChoice` seems fine to me, however for `Choice` I still think a `NamedTuple` is the way to go. `Choice` doesn't need any defaults, you can just make all four components mandatory, since it's for internal use only, so you also don't need to use any subclassing tricks, you can just directly use a plain `NamedTuple`.

```python
class Choice(NamedTuple):
    """
    A rendered option yielded by
    :meth:`SelectFieldBase.iter_choices` and
    :meth:`SelectFieldBase.iter_groups`.

    ``selected`` is computed against the field's current data. To
    declare options on a :class:`SelectField`, use
    :class:`SelectChoice` instead.

    :param value:
        The value that will be sent in the request.
    :param label:
        The label of the option.
    :param selected:
        Whether the option is currently selected. Set by ``iter_choices``;
        you rarely set this yourself.
    :param render_kw:
        A dict containing HTML attributes that will be rendered
        with the option.
    """

    value: str
    label: str
    selected: bool
    render_kw: dict
```

The main reason to use a `NamedTuple` for `Choice` is so that type checkers better understand them as structured types for the purposes of unpacking, you can't unpack a dataclass that has mixed field types and have type checkers know the type of each component. This preserves exact typing information and ergonomics for people that decide to keep their implementations the same, instead of switching to named attribute access, which can make the code longer.

I.e. compare

```python
for value, label, selected, render_kw in field.iter_choices():
    do_something_with(value, label, selected)
```

to

```python
for choice in field.iter_choices():
    do_something_with(choice.value, choice.label, choice.selected)
```

Neither code is obviously better than the other. Forcing people to switch their code to the latter, just to preserve good type checking experience seems like unnecessary downstream churn for no real benefit to anyone, when a `NamedTuple` would've worked just as well. It also avoids any potential risks for breakage in downstream code that expects a `tuple`, since they will still get one, since `NamedTuple` classes are subclasses of `tuple`.

azmeuk:
Honestly, about named tuples, I think this is good enough like that.
- Expecting people to change their code sounds acceptable for a minor version. That's what people need to do to get rid of deprecation warnings anyway.
- Sooner or later we will have native type annotations #618, and that compatibility layer is already complex enough. I am not sur it worth making design decisions to keep the DX unchanged in the meantime (as long as backward compatibility is maintained)
- The major dowstream projects don't break with this PR

Daverball:
It's not as much about breaking code in an obvious way, since you've increased runtime compatibility by implementing `__iter__` on `Choice`, but it will degrade what type checkers know about old code, that's making use of unpacking, so it's easier for regressions to sneak in, in the future, if the shape of `Choice` changes again. I also generally believe that API changes should be made extremely carefully and conservatively, especially on a minor version, that is especially true for any objects returned by your API. Any change that requires or strongly incentivizes downstream users to change their code immediately, needs to be justified extremely well, especially since people generally expect semantic version principles to be applied, so they may pin their dependencies to the next major version. If that contract gets violated it can very easily cause broken older releases, since they're not pinned to the last version they're compatible with.

Since `SelectChoice` is an entirely new way to define choices on the input side, it's fine to go with whatever layout you prefer, especially since people can wait with migrating to the new API until the next major/minor release and do the upgrade gradually, one module at a time, without needing to do it all at once. Encountering a `SelectChoice` object will also generally not come as a surprise, since you would've deliberately had to create them first, before you'd see them downstream.

The same is not true on the consumer side of `iter_choices` and `iter_groups`, if people suddenly receive objects of a completely different type, they will not get a deprecation warning, unless they specifically subclassed the `Select` widget and override `render_option`, they will either get a crash or worse some subtle regression, so I think it's important to strive for maximum compatibility here, which includes type checker compatibility, not just runtime compatibility, since they may very well be stuck on older versions of type stubs for a while, since larger API changes tend to take longer to arrive in typeshed, since they take longer to review. So those type stubs should remain correct for minor version bumps, as long as you don't start using the new API.

There's a very clear difference between the two types of changes, one is acceptable for a minor version, since it can be made gradually and at one's leisure. The other may very well require immediate action, not breaking the biggest downstream open source project test suites is not a good enough argument, especially when it's easy for you to mitigate the risks completely by switching to `NamedTuple`. Choosing a dataclass over a `NamedTuple` is taking an unnecessary risk.

TLDR: I personally do not see enough benefits to making `Choice` a dataclass, over a `NamedTuple` to justify the decision, so if `NamedTuple` provides superior backwards compatibility and type checking experience by giving people the freedom to continue unpacking choices instead of accessing attributes by name, it's just a total no brainer to me and I would once again very strongly urge you to change it. I've worked very hard to make `types-WTForms` a good experience with the proper balance between type coverage and user ergonomics and I'd hate for that experience to degrade, one subtle change at a time, in the interim as we transition to a natively typed API.

azmeuk:
Thank you for taking time to develop on this point.
I restored named tuples for Choice, and kept dataclasses for DataListChoice and SelectChoice. In the end this did not make as much difference as I expected, since we don't actually need default values for Choice because it is more an internal class.

I let you have a final view on this PR and tell me if you notice additional things.

src/wtforms/fields/choices.py
```diff
@@ -28,24 +33,62 @@ def coerce(v):
...
+        The ``<optgroup>`` HTML tag in which the option will be rendered.
+    """
+
     value: str
     label: str | None = None
-    render_kw: dict | None = None
-    _selected: bool = field(default=False, kw_only=True)
+    render_kw: dict = field(default_factory=dict)
+    optgroup: str | None = None
+
```
Daverball, line 85: This is pretty minor, but speaking from experience, it could get annoying that `label` is `str | None`, when, thanks to `__post_init__` it really should always be `str`. Once you actually start type checking this code base, this might result in a bunch of `assert SelectChoice.label is not None` or `type: ignore` comments, everywhere you use `SelectChoice.label`.

So it might be worth annotating the field as `str`, even though the default is `None`, this also gives us some additional safety that nobody accidentally sets this field back to `None` after the object has been initialized, which would be considered legal by type checkers, if the annotation is `str | None` instead of `None`.

Another possibility is storing `label` into an internal attribute and have a settable `property` as the public interface, but I don't think it's worth the extra complexity, it's better for us to have a couple of type ignore comments, to improve the ergonomics everywhere else that object is used. A property also seems less robust.

src/wtforms/fields/choices.py
```diff
@@ -60,48 +103,78 @@ def from_enum(cls, enum_cls, *, label=None):
...
-                stacklevel=2,
-            )
+            if len(input) == 2:
+                value, label = input
+                return cls(value=value, label=label, optgroup=optgroup)
+            if len(input) == 3:
+                value, label, render_kw = input
+                return cls(
+                    value=value, label=label, render_kw=render_kw, optgroup=optgroup
+                )
```
Daverball: This code seems unnecessary to me, `cls(*input, optgroup=optgroup)` will already do the correct thing for 2-element and 3-element tuples. It is a little bit more resilient towards changes in the order of fields in `SelectChoice`, but I don't think we should feel tempted to mess around with that in the first place, so it doesn't seem worth it.

It might be worth emitting a `ValueError` though, if the `tuple` is not of length 2 or 3, to give a slightly better error message, than what we would get from the call with too few or too many arguments otherwise.

src/wtforms/form.py
```diff
@@ -179,7 +201,7 @@ def __init__(cls, name, bases, attrs):
         cls._unbound_fields = None
         cls._wtforms_meta = None
 
-    def __call__(cls, *args, **kwargs):
+    def __call__(cls, *args, _parent_form=None, **kwargs):
```
Daverball: I like this solution, but some type checkers will probably get pretty confused if they see the `_parent_form` argument in a form constructor, passing this information in via the `meta` argument, would probably be a little less confusing.

Since `FormField` expects a `Form`, rather than `BaseForm`, this should be pretty straightforward, changing `_parent_form=self.form` to `meta={"_parent_form": self.form}`.

Daverball (commented):
Thanks for being so patient and receptive of my feedback. As far as I can tell this addresses any concerns I had and I didn't spot any other potential pitfalls this time around.

I spotted a couple of minor things I'd personally clean up, but they're low priority and won't affect end-users yet either way. So do with them what you want.

PR #926, opened 2026-06-01
Improve SelectField enum coerce

azmeuk:
Implementing suggestions from #922. No more implicit magic for `coerce=` with `Enum` values, `SelectChoice.coerce_by_name` is introduced to be used instead (`coerce=SelectChoice.coerce_by_name`). I am not delighted with the name but cannot find a better one.
/cc @miketheman @daverball

*edit: that damned Copilot review button...*

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -407,13 +407,27 @@ Choice Fields
             BLUE = 3
 
         class PaintForm(Form):
-            color = SelectField(choices=SelectChoice.from_enum(Color), coerce=Color)
+            color = SelectField(
+                choices=SelectChoice.from_enum(Color),
+                coerce=SelectChoice.coerce_by_name(Color),
+            )
 
-    :meth:`SelectChoice.from_enum` builds the option list from the Enum items;
-    the HTML ``value`` of each option is the item's ``name``. Passing the
-    Enum class itself as ``coerce`` installs the matching coercion, so
+    :meth:`SelectChoice.from_enum` builds the option list from the Enum
+    items; the HTML ``value`` of each option is the item's ``name``.
+    :meth:`SelectChoice.coerce_by_name` returns the matching ``coerce``
+    callable that resolves names back into members, so
     ``form.color.data`` is a ``Color`` item after submit. Pre-selecting
     works the usual way, with an Enum item: ``PaintForm(color=Color.RED)``.
+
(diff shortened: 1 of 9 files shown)


Daverball:
Looks good to me. I also can't think of a better name for the helper method, apart from maybe `coerce_enum_by_name`, to really drive home that this method is for enums, but the name also doesn't seem that bad to me in the first place, the only slightly awkward thing about it, is that `SelectChoice` may not be the most natural place to look for such a method, maybe it makes more sense to define it on `SelectField` itself?

docs/fields.rst
```diff
@@ -407,14 +407,28 @@ Choice Fields
...
     ``form.color.data`` is a ``Color`` item after submit. Pre-selecting
     works the usual way, with an Enum item: ``PaintForm(color=Color.RED)``.
 
+    If you'd rather submit ``item.value`` instead of ``item.name``,
+    pass the Enum class directly as ``coerce``; ``EnumCls(value)`` is
+    then used to resolve the submitted string::
+
+        class PaintForm(Form):
+            color = SelectField(
+                choices=[(m.value, m.name) for m in Color],
```
miketheman: Might as well suggest the new non-deprecated syntax?
```suggestion
                choices=[SelectChoice(m.value, m.name) for m in Color],
```

miketheman (commented):
I haven't tried the code yet, but thinking about naming, how might this look?

```python
class PaintForm(Form):
  color = SelectField(
    choices=SelectChoice.from_enum(Color),
    coerce=SelectChoice.enum_name(Color),
  )
```

That allows for some future `enum_value` that handles the other pattern.

The UX still "feels weird" overall - since I'm not asking to `coerce` one type into another type (like the `coerce_none` example in the docs), rather I'm asking the SelectField `choices` options of `(value, label, render_kw)` to be populated with some part of an enum, so maybe the overall `from_enum` could do it for me, instead of only allowing the `label` input to affect what's displayed, but also the value?

```python
class PaintForm(Form):
  color = SelectField(
    choices=SelectChoice.from_enum(
      Color,
      label=some_callable,
      value=some_other_callable),
  ),
```

azmeuk:
After giving some more thoughts I went to two independent `enum_choices` and `enum_coerce` methods with a `by` attribute, so either the enum label or the enum value can be used as form value.

PR #927, opened 2026-06-02
AI spam

aizu-m:
Both CSRF validators compare a secret token against attacker-supplied form data with `!=`, which short-circuits and leaks a byte-by-byte timing oracle on the expected token; switch `SessionCSRF` and the base `CSRF` to `hmac.compare_digest` (and stop treating a None/None pair as a match).

### DIFF
--- a/src/wtforms/csrf/core.py
+++ b/src/wtforms/csrf/core.py
@@ -1,3 +1,5 @@
+import hmac
+
 from wtforms.fields import HiddenField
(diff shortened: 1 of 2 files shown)


ThiefMaster:
This is not a problem for CSRF tokens at all. You can't find out the CSRF token for another user's session that way, so all you could do is trying it for your own session - but that's pointless, because you already have access to it.

PR #934, opened 2026-08-02
AI junk

### LINKED ISSUE #865: Default type for `Field.errors` of `tuple()` is immediately overwritten
The initial data type for `Field.errors` is set to a tuple:
https://github.com/pallets-eco/wtforms/blob/9cc223ad2fabeabd8d89021ea962b106d1502b6c/src/wtforms/fields/core.py#L19

which is reassigned to a `list` if there's any errors:
https://github.com/pallets-eco/wtforms/blob/9cc223ad2fabeabd8d89021ea962b106d1502b6c/src/wtforms/fields/core.py#L215

which causes type checkers like mypy to be sad when you use the field in your own code directly.

### Actual Behavior

in `repro.py`:

```python
import wtforms

class MyForm(wtforms.Form):
    name = wtforms.StringField()

    def validate(self, extra_validators=None, *args, **kwargs):
        super().validate(extra_validators, *args, **kwargs)

        if self.name.data == 'repro':
            # note the use of `append()`, unavailable on a tuple.
            self.name.errors.append('Name cannot be "repro"')
```

```shellsession
$ mypy --check-untyped-defs repro.py
repro.py:10: error: "Sequence[str]" has no attribute "append"  [attr-defined]
Found 1 error in 1 file (checked 1 source file)
```

### Expected Behavior

```shellsession
$ mypy --check-untyped-defs repro.py
Success: no issues found in 1 source file
```

### Environment

* Python version: 3.12
* wtforms version: 3.2.1
* mypy version: 1.12.1

Yes, this specific example could be solved another way with `validate_name`, but that's not the point. The point is that the error's initial datatype is incorrect.
Everywhere it's used, it's treated as a list.

I'd be happy to send a PR changing to an empty list, but I didn't know if there was some other reason to have the data type as a tuple initially.

Sanjays2402:
Closes #865

`Field.errors` and `Field.process_errors` are class-level tuples until `Field.process()` runs, and `errors` stays a tuple until `Field.validate()` replaces it with a list. So appending to `field.errors` from a `Form.validate()` override raises `AttributeError`, and type checkers infer `Sequence[str]` instead of `list[str]`.

Link to any relevant issues or pull requests: #865

`Field.process()` now resets both attributes to empty lists, matching how they are used everywhere else in the codebase (`validate`, `FieldList.validate`, the `Optional`/`InputRequired` validators all treat them as lists). `FormField` still overrides `errors` as a property, so it is unaffected.

Commit checklist:

* [x] add tests that fail without the patch — `test_errors_are_a_list_before_validation` in `tests/fields/test_field.py` fails on `main` and passes with the patch
* [x] ensure all tests pass with `pytest` — same results before and after (the only local failures are pre-existing, from missing compiled translation catalogs)
* [x] add documentation to the relevant docstrings or pages — no public API change to document
* [x] add `versionadded`/`versionchanged` directives — n/a, no signature change
* [x] add a changelog entry

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.


davidism:
https://palletsprojects.com/contributing/llm-ai

